PR #195, opened 2016-07-01
Allow start and end dates to be passed into the share harvest utility

erinspace:
add two option args, --start and --end, that will allow you to pass in start and end times to the harvest utility

Usage:
```./bin/share harvest org.arxiv --limit=10 --start=2016-01-01 --end=2016-01-10```


chrisseto:
@erinspace could you update this PR when you get the chance?

erinspace:
@chrisseto PR Updated!

PR #196, opened 2016-07-01
[WIP] Update the OAI Normalizer

leb2dg:
* add example records
* update bhl
* comment on OAI normalizer things we need to support


leb2dg:
Closed in favor of #202. Also, this code is already in the code base.

PR #199, opened 2016-07-04
Update BioRxiv normalizer

leb2dg:
* switch () and []
* use new `Maybe` and `Map` syntax
* specify imports
* add publisher
* update doi field
* add normalize/utils.py


providers/org/biorxiv/normalizer.py
```diff
@@ -1,22 +1,37 @@
...
-    contributors = ctx.item('dc:creator')['*']
-    published = ctx.item('dc:date')
-    doi = ctx.item('dc:identifier')
+    contributors = links.Map(links.Delegate(Contributor), ctx.item['dc:creator'])
+    published = ctx.item['dc:date']
+    publishers = links.Map(
+        links.Delegate(Association.using(entity=links.Delegate(Publisher))),
+        ctx.item['dc:publisher']
+    )
+    # doi = ctx.item['dc:identifier']
```
leb2dg: @chrisseto, what should be done with the doi field?

chrisseto: Create a parser for `Link` and `ThroughLink` 
https://github.com/CenterForOpenScience/SHARE/blob/develop/share/models/meta.py#L42
I'll make the `ThroughLink` model in a second

providers/org/biorxiv/normalizer.py
```diff
@@ -1,22 +1,51 @@
...
+
+
+class Association(Parser):
+    pass
+
+
+class Link(Parser):
+    url = links.RunPython('format_doi', ctx)
+    # identifier will always be DOI
+    type = 'doi'
```
chrisseto: did you decide what to do about this?

chrisseto:
pass done. Couple comments. otherwise lgtm

providers/org/biorxiv/normalizer.py
leb2dg: For this provider it identifier is only used for DOIs. Is always setting `type` to 'doi' adequate?

leb2dg:
@chrisseto, StaticLink has been updated

PR #201, opened 2016-07-05
Update Arxiv Normalizer and BioRxiv import

leb2dg:
* add doi to Arxiv normalizer
* import tools instead of links for Arxiv and BioRxiv


leb2dg:
@chrisseto, this is ready for review 💸

PR #205, opened 2016-07-06
Harvester and Normalizer for OSF Registrations

erinspace:
<strike>Note: Blocked by an OSF API bug, should be merged and deployed soon. Will update when that happens!</strike>


providers/io/osf/registrations/normalizer.py
```diff
@@ -0,0 +1,119 @@
...
+class Tag(Parser):
+    name = ctx
+
+
+class ThroughTags(Parser):
+    tag = Delegate(Tag, ctx)
+
+
+class Institution(Parser):
+    name = Maybe(ctx, 'attributes.name')
```
chrisseto: This probably doesn't work

PR #209, opened 2016-07-07
update arxiv normalizer with new model fields

leb2dg:


leb2dg:
@chrisseto, this is ready 👍

PR #217, opened 2016-07-07
Add Springer harvester/normalizer, Remove Biomedcentral

erinspace:
For awhile now, Springer and BioMedCentral have both used the exact same Springer API. They were separate in scrapi for legacy reasons, but can now be combined as there is no good way to separate the two without doing post processing.

Springer harvester/initial normalizer written by @caseyrollins 💃


chrisseto:
Closing as we need to keep biomedcentral around for the migration.

PR #218, opened 2016-07-07
USGS Harvester & Normalizer

caseyrollins:


caseyrollins:
@chrisseto 🍭

providers/gov/usgs/normalizer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Person(Parser):
+    given_name = Maybe(ctx, 'given')
+    family_name = ctx.family
+
+
+class Contributor(Parser):
+    order_cited = ctx('index')
+    cited_name = Join(Concat(Maybe(ctx, 'given'), ctx.family), ' ')
```
chrisseto: `joiner=' '`?

providers/gov/usgs/normalizer.py
```diff
@@ -0,0 +1,82 @@
...
+class ThroughLinks(Parser):
+    link = Delegate(Link, ctx)
+
+
+class Publisher(Parser):
+    name = ctx
+
+
+class Association(Parser):
+    pass
```
chrisseto: do da thing

caseyrollins:
@chrisseto updated 🍸

providers/gov/usgs/harvester.py
```diff
@@ -1,5 +1,33 @@
...
 
 class USGSHarvester(Harvester):
-    pass
+    url = 'https://pubs.er.usgs.gov/pubs-services/publication'
+
+    def do_harvest(self, start_date, end_date):
+        end_date = end_date.date()
+        start_date = start_date.date()
+
+        days_back = (end_date - start_date).days
```
chrisseto: This a bit odd to do, if I run this harvester from 2014 - 2015 it would end up collecting data from 2016.
There should probably be a check that start date is today.

providers/gov/usgs/harvester.py
```diff
@@ -1,5 +1,33 @@
...
+            'mod_x_days': days_back,
+        }).url)
+
+        records = resp.json()['records']
+        page = 1
+
+        while records:
+            for record in records:
+                record_id = record['id']
+                yield(record_id, record)
```
chrisseto: space between `yield` and `(`

caseyrollins:
@chrisseto Updated this harvester ☕

providers/gov/usgs/harvester.py
```diff
@@ -1,5 +1,42 @@
...
+        end_date = end_date.date()
+        start_date = start_date.date()
+
+        end_days_back = (today - end_date).days
+        start_days_back = (today - start_date).days
+
+        # The USGS API does not support date ranges
+        for days_back in range(end_days_back, start_days_back):
+            page = 1
+            resp = self.requests.get(furl(self.url).set(query_params={
```
chrisseto: Sorry to be nitpicky here, but could you refactor this to be a bit more DRY?
It's kinda sucky to do because of python's lack of do while loops but it leads to better maintainability. 
[This SO post](https://stackoverflow.com/questions/743164/emulate-a-do-while-loop-in-python) is the best way I've come across to get psuedo do whiles

providers/gov/usgs/normalizer.py
```diff
@@ -0,0 +1,82 @@
...
+
+    def format_doi_as_url(self, doi):
+        return format_doi_as_url(self, doi)
+
+    def format_usgs_id_as_url(self, id):
+        return 'https://pubs.er.usgs.gov/publication/{}'.format(id)
+
+    def get_links(self, links):
+        for link in links:
+            return link.get('url')
```
chrisseto: This will only return the first link, is that intended?

providers/gov/usgs/normalizer.py
```diff
@@ -0,0 +1,82 @@
...
+
+
+class Link(Parser):
+    url = ctx
+    type = RunPython('get_link_type', ctx)
+
+    def get_link_type(self, link):
+        if 'dx.doi.org' in link:
+            return 'doi'
+        elif 'usgs.gov' in link:
```
chrisseto: If you have `return`s in your `if` statements don't use `else` or `elif`

Excerpt from [Effective Go](https://golang.org/doc/effective_go.html#if)

> In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

caseyrollins:
@chrisseto 🔙

providers/gov/usgs/normalizer.py
```diff
@@ -0,0 +1,81 @@
...
+    name = ctx
+
+
+class Association(Parser):
+    entity = Delegate(Publisher, ctx)
+
+
+class CreativeWork(Parser):
+
+    def format_doi_as_url(self, doi):
```
chrisseto: Methods should be defined below class level variables.

```python

class MyClass
    class_level_var = 1

    @classmethod
    def my_class_method(cls):
        pass

    @property
    def my_property(self):
        return 'things'

    def __init__(self, *args):
        self.instance_var = args

    def my_method(self, arg):
        pass

```

caseyrollins:
@chrisseto 🌜

providers/gov/usgs/harvester.py
```diff
@@ -1,5 +1,46 @@
...
+            page_size = 100
+
+            while True:
+                resp = self.requests.get(furl(self.url).set(query_params={
+                    'mod_x_days': days_back + 1,
+                    'page_number': page,
+                    'page_size': page_size
+                }).url)
+
+                print(furl(self.url).set(query_params={
```
chrisseto: Please remove this print

chrisseto:
@caseyrollins🌛

caseyrollins:
@chrisseto 🌜

PR #219, opened 2016-07-08
LWBIN Harvester & Normalizer

caseyrollins:


caseyrollins:
@chrisseto 👠

providers/ca/lwbin/harvester.py
```diff
@@ -2,4 +2,13 @@
...
 class LWBINHarvester(Harvester):
-    pass
+    # LWBIN does not have an HTTPS URL
+    url = 'http://130.179.67.140/api/3/action/current_package_list_with_resources'
+
+    def do_harvest(self, start_date, end_date):
+        # Searching by time is not permitted by the LWBIN CKAN API. All records must be scanned each time.
+        response = self.requests.get(self.url)
+        records = response.json()['result']
+        for record in records:
```
chrisseto: How many records are there are we talking about here?

providers/ca/lwbin/harvester.py
```diff
@@ -2,4 +2,13 @@
...
+            record_id = record['id']
+            yield(record_id, record)
```
chrisseto: space between `yield` and `(`

providers/ca/lwbin/harvester.py
caseyrollins: ~60

caseyrollins:
@chrisseto updated 🎱

providers/ca/lwbin/harvester.py
chrisseto: Could you add in pagination support?
> Return a list of the site's datasets (packages) and their resources. The list is sorted most-recently-modified first. :param limit: if given, the list of datasets will be broken into pages of at most ``limit`` datasets per page and only one page will be returned at a time (optional) :type limit: int :param offset: when ``limit`` is given, the offset to start returning packages from :type offset: int :param page: when ``limit`` is given, which page to return, Deprecated use ``offset`` :type page: int :rtype: list of dictionaries 

It looks like you can provide `limit` and `page`. if `len(results) == limit`  then there is another page.

caseyrollins:
@chrisseto 🌜

PR #221, opened 2016-07-08
ParseLanguage Link

caseyrollins:
- Parses languages to their ISO language codes.
- Update language field for SciTech normalizer and OAI normalizer.


caseyrollins:
@chrisseto ISO codes 🌸

PR #223, opened 2016-07-08
NODC OAI

caseyrollins:
NODC is actually an OAI provider.


chrisseto:
@caseyrollins Could you remove the original harvester stub?

caseyrollins:
@chrisseto Removed 💯

PR #227, opened 2016-07-09
Adding indexes to people

cwisecarver:
Adding indexes to people fields to try and speed up disambiguation.


cwisecarver:
Checked the field lengths on staging for these fields. GTG

PR #231, opened 2016-07-10
Add 'date' field to creative work search

aaxelb:
Single date field for search filtering. Prefer `date_published`, if it exists, or fall back to `date_updated`, or `date_created` if all else fails.


erinspace:
I have a feeling people might not love a generalized date field as this is losing some information - would there be a way to also include the other date information as well as the generalized one? So maybe:

```
dates: {
    date_published: 2016-06-01,
    date_created: null,
    general_date: 2016-01-01 
}
```

erinspace:
Oh nevermind I totally missed that they were already included above, disregard!!

PR #241, opened 2016-07-10
Added examples of Venue classes + trivial style edits

bnosek:
Followed the Awards example to create Venue example so that there is no ambiguity.  Followed format precisely - including use of "venue" and "venues" following the different use of "award" and "awards".  This change made assuming the earlier instruction "syntax follows the awards example" is literal.  If this is correct, I can repeat addition for funders, institutions, and organizations so that all are explicit.

### DIFF
--- a/docs/share_models.rst
+++ b/docs/share_models.rst
@@ -3,7 +3,7 @@ Share Models
 
 Model Descriptions
 ------------------
-The descriptions of the models used in SHARE will be useful when writing the normalizer for a new provider.
+SHARE model descriptions will be useful when writing normalizers for new providers.
 See existing provider normalizers for more detailed examples.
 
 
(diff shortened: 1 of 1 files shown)


docs/share_models.rst
```diff
@@ -38,7 +38,18 @@ Creative Work
         class CreativeWork:
             awards = Map(Delegate(ThroughAwards), ctx.<awards>)
   - venues
-   - A list of venues associated with the work, passed to the ``Venue`` class via the ``ThroughVenues`` class (syntax follows the ``awards`` example above).
+   - A list of venues associated with the work, passed to the ``Venue`` class via the ``ThroughVenues`` class::
+   
+        class Venue:
+            description = ctx.<venue_description>
```
sloria: Since this looks to be a runnable code example, the angle-brackets should be removed to make this valid Python. I'll do this when I merge.

sloria:
Thanks!

PR #252, opened 2016-07-11
Hotfix/entity autocomplete

aaxelb:
All entities in the autocomplete index have `"text": "Publisher object"` (or whatever the type) instead of their name. Tags were fine because they have a useful `__str__`.


icereval:
Fixed @ https://github.com/CenterForOpenScience/SHARE/commit/8f47ffdac2f3d6f5235c38988d3c8dea10d883a6

PR #269, opened 2016-07-11
Normalize extra data from PLOS

### LINKED ISSUE #250: [normalizer] Update PLOS Normalizer with Extra Data
The PLOS normalizer should have an ```Extra``` class within ```CreativeWork``` for some of the data that is not currently being normalized. The fields currently being ignored are noted in ```otherProperties``` in the PLOS scehma [here](https://github.com/CenterForOpenScience/scrapi/blob/develop/scrapi/harvesters/plos.py).

caseyrollins:
- Add Extra class to PLOS normalizer.
- Normalize publisher information.
- Normalize DOI as a link.

Closes #250

### DIFF
--- a/providers/org/plos/normalizer.py
+++ b/providers/org/plos/normalizer.py
@@ -17,7 +17,12 @@ class Contributor(Parser):
 
 class Link(Parser):
     url = ctx
-    type = Static('doi')
+    type = RunPython('get_link_type', ctx)
+
+    def get_link_type(self, link):
+        if 'dx.doi.org' in link:
+            return 'doi'
+        return 'eissn'
 
 
 class ThroughLinks(Parser):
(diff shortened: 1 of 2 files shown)


providers/org/plos/normalizer.py
```diff
@@ -14,12 +15,48 @@ class Contributor(Parser):
...
+        RunPython(
+            'format_doi_as_url',
+            XPath(ctx, "str[@name='id']").str['#text']
+        )
+    )
     date_published = ParseDate(XPath(ctx, "date[@name='publication_date']").date['#text'])
-    # doi = ctx.xpath("str[@name='id']").str['#text']
+
+    class Extra:
+        eissn = XPath(ctx, "str[@name='eissn']").str['#text']
```
chrisseto: I think this should be a type of link. Eissns don't have a way of being expressed as UR[I|L|N]s currently, which makes this a bit odd.
Let's go with `eissn://<value>` and type as `esissn`

caseyrollins:
@chrisseto EISSNs are now links 🌵

providers/org/plos/normalizer.py
```diff
@@ -14,12 +15,53 @@ class Contributor(Parser):
...
 
 
+class Link(Parser):
+    url = ctx
+    type = RunPython('get_link_type', ctx)
+
+    def get_link_type(self, link):
+        if 'dx.doi.org' in link:
+            return 'doi'
+        return 'eissn'
```
chrisseto: Could you make this a regex match or something a little bit less brittle?

chrisseto:
Merged Via https://github.com/CenterForOpenScience/SHARE/commit/b59a4d77a9e4e7141ac2ecfb4b7060fffc2507c4

Issue #257, opened 2016-07-11
[providers] Add e-LiS Harvester & Normalizer


caseyrollins:

SSJohns:
Looking into this one.

PR #276, opened 2016-07-12
Feature/e lis harvester normalizer

SSJohns:
## Purpose

Adds E-LIS to the webarchives that are covered and scraped for.

## Changes

Added scarper for the E-LIS, and as it is a OAI it uses the default scarper and harvester for those libraries that are OAI enabled. There was also a case where the scaper could not handle multiple languages and this was taken care of in an update to the way languages are handled. Now the first language in the list is just passed to the Parse Language.

## Side effects

Will now check through all the languages and only grab the first one in the case of multiple languages

## Ticket

#257


providers/org/elis/normalizer.py
```diff
@@ -0,0 +1,3 @@
+from share.normalize.oai import OAICreativeWork
```
chrisseto: Please delete this file, it is not required.

share/normalize/oai.py
```diff
@@ -120,7 +120,15 @@ class OAICreativeWork(Parser):
 
     rights = tools.Join(tools.Maybe(tools.Maybe(ctx.record, 'metadata')['oai_dc:dc'], 'dc:rights'))
 
-    language = tools.ParseLanguage(tools.Maybe(tools.Maybe(ctx.record, 'metadata')['oai_dc:dc'], 'dc:language'))
```
chrisseto: Could you use the `TryLink` add in [this PR](https://github.com/CenterForOpenScience/SHARE/pull/283)
You'll be able to do something like `tools.Try(ctx.record[metadata][oai_dc:dc][dc:language][0])`

SSJohns: Done

Issue #184, opened 2016-05-19
Let's harvest from RePEc


JudyRuttenberg:
http://repec.org/

DanielSBrown:
Looking into this one

PR #277, opened 2016-07-12
Added MPRA (Munich Personal RePEc Archive) harvester and its migrations

DanielSBrown:
After trying to just make a harvester for just the general RePEc website, I ran into problems.  There was no real data to be harvested from it; instead it had many of its harvestable files on the MPRA.  So, I made a harvester for that instead.  

# **Purpose**

- Add harvester for files on the MPRA

- Add relevant migrations too

# **Changes**

- Make a ``providers/org/mpra`` directory housing the ``apps.py`` and ``__init__.py``

- Make the migrations necessary in the ``0001_initial.py`` and ``__init__.py`` in the migrations directory

# **Side Effects**

- Since it does not get all of RePEc, may need to include other harvesters for the subdivisions of RePEc (i.e. EconPapers, EDIRC)

#184


providers/org/mpra/apps.py
```diff
@@ -0,0 +1,10 @@
+from share.provider import OAIProviderAppConfig
+
+
+class AppConfig(OAIProviderAppConfig):
+    name = 'providers.org.mpra'
+    version = '0.0.1'
+    title = 'mpra'
+    long_title = 'Munich Personal RePEc Archive'
+    home_page = 'http://mpra.ub.uni-muenchen.de'
+    url = 'http://mpra.ub.uni-muenchen.de/perl/oai2'
```
chrisseto: Please add a new line at the end of this file

chrisseto: @erinspace does this URL require additional url params, such as metadata namespace?

chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/commit/844ce765976318a4e9c6754331b17a57b4c3e02b

PR #280, opened 2016-07-13
Datacite(oai_datacite) Harvester & Normalizer

leb2dg:
### Changes

* update RunPython to take function or function name
* used TryLink
    ```
    ('relatedIdentifiers', None),
    ```
* added `THE_REGEX` in normalizer which finds "the " or " the "
* added `DATE_REGEX` in normalizer which does not match `a-zA-Z`

    >
    YYYY,YYYY‐MM‐DD, YYYY‐
    MM‐DDThh:mm:ssTZD or any
    other format or level of
    granularity described in
    W3CDTF.17 Use RKMS‐
    ISO860118 standard for
    depicting date ranges.
    Example: 2004‐03‐02/2005‐
    06‐02
    >

    ```
    ('publicationYear', '0000'),
     ('dates',
            OrderedDict([('date', OrderedDict([('@dateType', 'Available'),
             ('#text', 'uuuu')]))])),
    ```


providers/org/datacite/oai/apps.py
```diff
@@ -0,0 +1,10 @@
+from share.provider import OAIProviderAppConfig
```
chrisseto: Is there any reason to keep the old OAI harvester around? It seems like the new namespace would be a superset of the oai namespace?

share/normalize/links.py
```diff
@@ -368,7 +409,37 @@ def execute(self, obj):
...
+        if isinstance(obj, str):
+            return obj
+        elif isinstance(obj, dict):
+            if '#text' in obj:
+                return obj['#text']
+            raise Exception('#text is not in {}'.format(obj))
+        elif isinstance(obj, list):
+            for item in obj:
+                if isinstance(item, str):
+                    return item
```
chrisseto: This will return the first string found in a list. Is that the expected functionality?

share/normalize/links.py
```diff
@@ -368,7 +409,37 @@ def execute(self, obj):
...
+            raise logger.warning('No value in list {} is a string.'.format(obj))
```
chrisseto: ` raise Exception('No value in list {} is a string.'.format(obj))`

share/normalize/links.py
leb2dg: Yeah, it is used for the title field where we really only want one string.

share/normalize/links.py
leb2dg: Unfortunately, raising an exception here will cause some issues because if the values are highly nested then we don't want them. For instance, titles can be abstracts, alternate, or translated but the main title is usually (but not always) just a string in the list.

leb2dg: I think this is actually a better example:`tools.Try(tools.Text(ctx.nameIdentifier))`

`nameIdentifier` won't always be there but we don't wan't everything to break if it's not there.

providers/org/datacite/normalizer.py
```diff
@@ -0,0 +1,523 @@
...
+
+
+class RelatedLink(Parser):
+    schema = 'Link'
+
+    url = tools.Text(ctx)
+    type = tools.RunPython('lower', tools.Try(ctx['@relatedIdentifierType']))
+
+    def lower(self, type):
+        if type:
```
chrisseto: Unless `'@relatedIdentifierType'` could contain a `None` value, type will always be defined here. `Try` will not execute the `RunPython` if it fails to find `'@relatedIdentifierType'`.

leb2dg:
@chrisseto, updated 🔆

providers/org/datacite/normalizer.py
```diff
@@ -0,0 +1,585 @@
...
+from share.normalize.utils import format_doi_as_url
+
+
+logger = logging.getLogger(__name__)
+
+THE_REGEX = re.compile(r'(^the\s|\sthe\s)')
+DATE_REGEX = re.compile(r'(^[^a-zA-Z]+$)')
+
+
+def force_text(data):
```
chrisseto: Do you know how this will deal with `None`s? There's another version in the oai harvester that blew up in the face of `None`.

Here's the offending tag:
```xml
<dc:title/>
```

leb2dg: I don't why `None` is in the metadata. I haven't seen `None` in any required field for the `oai_datacite` prefix though (title is one of the required fields).

providers/org/datacite/normalizer.py
```diff
@@ -0,0 +1,585 @@
...
+
+
+class ContributorPerson(Parser):
+    schema = 'Person'
+
+    suffix = tools.ParseName(ctx.contributorName).suffix
+    family_name = tools.ParseName(ctx.contributorName).last
+    given_name = tools.ParseName(ctx.contributorName).first
+    additional_name = tools.ParseName(ctx.contributorName).middle
+    identifiers = tools.Map(tools.Delegate(ThroughIdentifiers), tools.Concat(tools.Try(ctx, 'nameIdentifier')))
```
chrisseto: Do you need the `Concat` here?

providers/org/datacite/normalizer.py
```diff
@@ -0,0 +1,585 @@
...
+        except KeyError:
+            return None
+
+    def list_in_string(self, string, list_):
+        for word in list_:
+            if isinstance(word, str):
+                if word in string.lower():
+                    return True
+            else:
+                if word.search(string):
```
chrisseto: Where is search coming from?

leb2dg: `word` in that instance is a regex

providers/org/datacite/normalizer.py
leb2dg: If there is only one `nameIdentifier` then `tools.Try(ctx, 'nameIdentifier')` returns a string.

project/settings.py
```diff
@@ -174,6 +174,7 @@
     'providers.org.cogprints',
     'providers.org.crossref',
     'providers.org.datacite',
+    'providers.org.datacite.oai.apps.AppConfig',
```
chrisseto: You should be able to just use `providers.org.oai`

PR #281, opened 2016-07-15
Feature/funders awards update trigger migrations

cwisecarver:
@chrisseto This should add funders -> awards. 

It also changes `maketriggermigrations`, by default, create a new migration that removes the current triggers and stored procs and replaces them with updated ones based on the fields that are currently there. If you pass `--initial` to it then it will assume you're starting from scratch. Since the `DROP` is included in the command it should be relatively fool-proof. We might get some duplicative migrations if people run `maketriggermigrations` when they shouldn't.


share/models/meta.py
```diff
@@ -26,6 +26,7 @@ class Award(ShareObject):
     award = ShareURLField(blank=True)
     description = models.TextField(blank=True)
     url = ShareURLField(blank=True)
+    funder = ShareManyToManyField('Funder', through='ThroughAwardFunders')
```
chrisseto: Thinking out loud here, should this just be a generic sponsor that could be any entity?

cwisecarver: I can't really see any reason not to. I was just doing what the group suggested. Maybe an award could be funded by an Organization or Publication. I don't know. They just mentioned Funder.

chrisseto: Let's go with the more generic option for now.

cwisecarver: I see what you're doing. You're trying to pull me back into SHARE.

chrisseto: I or someone else could make the change if you would like

cwisecarver: HyperTerm is really pretty neat. Especially with wow hyperpower mode. But tmux doesn't work in it.

cwisecarver:
@chrisseto It's pointing at entities now. I also added entities to the serializer.

PR #283, opened 2016-07-18
[Feature] TryLink and use run instead of execute

leb2dg:
### Changes
* TryLink
    * similar to Maybe
* DateParser
   * handle being passed `None`


leb2dg:
@chrisseto, updated!

Issue #291, opened 2016-07-19
Favicons for providers


chrisseto:
Much like scrapi we should have favicons for each provider and then use django to put them in a static directory.

SSJohns:
For this do you mean to grab the icons from the specific directory and then place them all in one place? So the directory will have the favicons and I just need to pull them from there right?

PR #311, opened 2016-07-20
feature/adding-favicons-to-provider-directories

SSJohns:
Will pull together all the files from the static directories in the system and put them in one place.

Should not have any ill effects as it is not modifying current files. Although access of the favicons will greatly depend on the file structure provided by the script in the other PR.
#291


SSJohns:
@chrisseto Ready to review.

chrisseto:
You should be able to just but files in `/static/img/..` for each provider. Django *should* be namespacing static files.
The section right above [this one](https://docs.djangoproject.com/en/1.9/howto/static-files/#serving-static-files-during-development) has some details on how to do it.

SSJohns:
I'm sorry what do you mean? When you run `python manage.py collectstatic` it grabs all of these files and puts them into the root directory that you are talking about. That's why I had the script, it grabbed all the favicons and put them into the folders and the above command relocated them all to one place.

chrisseto:
Right now the all the static folder structures look like `/static/ca.lwbin/...`. Using django there should be a way to have each individual provider just store files like `/static/img/...` and have `collectstatic` namespace them, so the final static directory will look like `/static/ca.lwbin/...`

Could you also remove the favicon folder? `img/favicon.ico` is fine.

SSJohns:
You asked me to put in the favicon folder.... So let me get this straight before I go ahead.

You want in each folder named `/provider/edu/calpoly` you want a folder called `static/img` and the favicon.ico in that folder. Then when you pull it together with Django you want Django to name it `static/<providerappname>/img/favicon.ico`. Is that right or am I missing something?

chrisseto:
Spot on

SSJohns:
I think my folder structure the first time was right, there is no django way to do autonamespacing. The tutorials suggest to make another directory in your app like `/static/<providerappname>/img/etc.ico`.

chrisseto:
Whoops, looks like I misread the documentation. It's ridiculous that they don't do that automatically.

SSJohns:
Okay that should be everything @ chrisseto

chrisseto:
Beautiful, sorry for the confusion!

PR #315, opened 2016-07-22
Feature/repec provider

SSJohns:
Repec provider added. Takes some time to pull from their database as their servers are generally bad

#303


chrisseto:
@SSJohns Looks great, could you use `org.repec` rather than `org.openlib.repec`.

we're trying to keep everything at a single level of nesting.

SSJohns:
@chrisseto Done. It should all be right now. Still take a long time to harvest from these guys though.

providers/org/repec/apps.py
```diff
@@ -0,0 +1,12 @@
+from share.provider import OAIProviderAppConfig
+
+
+class AppConfig(OAIProviderAppConfig):
+    name = 'providers.org.repec'
+    version = '0.0.1'
+    title = 'repec'
+    long_title = 'Research Papers in Economics'
+    home_page = 'http://repec.org/'
+    url = 'http://oai.repec.openlib.org'
```
chrisseto: `http://oai.repec.openlib.org` redirects to `http://oai.repec.org/`

SSJohns:
@chrisseto

PR #316, opened 2016-07-22
Add/update favicon files

leb2dg:
### Changes
* add favicons for elis and mpra
* rename uow and pcurio favicon files
* don't ignore the favicon static directories


.gitignore
```diff
@@ -139,6 +139,22 @@ uk.*/
 uk.*/
 za.*/
 
+!providers/*/*/*/be.*/
```
chrisseto: You can also just add a `/` before all the `<tld>.*/`s in the above block.

The `/` limits the pattern to only matching files/folders in the root of the git directory.

PR #279, opened 2016-07-13
PeerJ harvester

zamattiac:
Submitting incomplete work, as requested.

### Completed:
- The harvester has been built and tested
- Various minutiae associated with a new provider taken care of

## TODO:
- A normalizer, maybe.

Check the \__init__ file to find the example record. Many of its fields should be standard.

PR #334, opened 2016-07-28
Feature/peerj

SSJohns:
Added the mormalizer to the existing harvester made by red-head Matt(who I can't link to for some reason). Detailed in #279 . #298 

Only question has a doi in the format: "doi": "10.7717/peerj.2215",

Should I be converting this to a link before storing it, or is it fine to just be stored as it.


leb2dg:
The doi should be converted to a link since it will be stored in the link class. You should be able to use [format_doi_as_url](https://github.com/CenterForOpenScience/SHARE/blob/develop/share/normalize/utils.py#L4). Datacite's normalizer uses it if you want to look at an example.

providers/com/peerj/apps.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class AppConfig(ProviderAppConfig):
+    name = 'providers.com.peerj'
+    version = '0.0.1'
+    title = 'peerj'
+    long_title = 'PeerJ'
+    home_page = 'https://peerj.com/articles/'
+    harvester = PeerJHarvester
+    # normalizer = PeerJNormalizer
```
chrisseto: Newline, please.

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,123 @@
...
+class Publisher(Parser):
+    name = ctx
+
+class Institution(Parser):
+    name = RunPython('get_author_institute', ctx)
+
+    def get_author_institute(self, context):
+        # read into a set and passed back to erase duplicates
+        if 'author_institution' in context:
+            inst_set = set(context['author_institution'])
```
chrisseto: You're losing the order here.

```python
In [5]: x = ['string1', 'another string', 'repeat', 'repeat', 'another string']

In [6]: list(set(x))
Out[6]: ['another string', 'repeat', 'string1']
```

2 potentials solutions are:

Create a dictionary that is index: value
Reverse it
Order the keys by value
```python
 x = ['string1', 'another string', 'repeat', 'repeat', 'another string']

x = dict(x)

x = reverse(x)

sorted(x.items())
```

Create a set and then remove values as you iterate over it
```python
x = ['string1', 'another string', 'repeat', 'repeat', 'another string']

other_x = set(x)

[other_x.pop(s) for s in x if s in other_x]
```

SSJohns:
@leb2dg Why do I need to pass self to format_doi_as_url? Is that important or what does it do.

leb2dg:
Unsure, you could try removing it as a parameter and see what happens. If you do remove it though make sure you update all of the normalizers that use that function.

providers/com/peerj/harvester.py
```diff
@@ -0,0 +1,39 @@
...
+                break
+
+            if arrow.get(record['date']) > end_date:
+                continue
+
+            yield (record['identifiers']['peerj'], record)
+
+        else:
+            if next_page:
+                self.fetch_records(next_page, start_date, end_date)
```
chrisseto: This will not work as expected. This call is doing nothing right now. It might work as a `yield from` but double check.

SSJohns: What is this yield from? Are there any examples, I only found it used once in the code base.

chrisseto: [Generator delegation](https://docs.python.org/3/whatsnew/3.3.html#pep-380)

SSJohns:
@chrisseto

providers/com/peerj/apps.py
SSJohns: Hey for this, do I need the `normalizer = PeerJNormalizer`. It seems to work fine without that line.

providers/com/peerj/harvester.py
```diff
@@ -0,0 +1,56 @@
...
+        records = self.requests.get(url).json()
+        next_page = records['_links']
+        for record in records['_items']:
+            if arrow.get(record['date']) < start_date:
+                return
+
+            if arrow.get(record['date']) > end_date:
+                continue
+
+            yield (preprint+record['identifiers']['peerj'], record)
```
chrisseto: Is it possible for there to be a conflict between preprints and non-preprints?

providers/com/peerj/harvester.py
```diff
@@ -0,0 +1,56 @@
...
+
+        if ('next' in next_page):
```
chrisseto: No parens

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,125 @@
...
+
+class Institution(Parser):
+    name = RunPython('get_author_institute', ctx)
+
+    def get_author_institute(self, context):
+        # read into a set while preserving order and passed back to erase duplicates
+        seen = set();
+        if 'author_institution' in context:
+            return [x for x in context['author_institution'] if x not in seen and not seen.add(x)]
+        else:
```
chrisseto: Don't use `else` when you're returning out of the if statement
https://golang.org/doc/effective_go.html#if

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,125 @@
...
+        Map(Delegate(ThroughLinks), ctx.fulltext_html_url)
+    )
+    publishers = Map(
+        Delegate(Association.using(entity=Delegate(Publisher))),
+        ctx.publisher
+    )
+    institutions = Map(
+            Delegate(Association.using(entity=Delegate(Institution))),
+            ctx
+        )
```
chrisseto: Fix the spacing here

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,125 @@
...
+        journal_abbrev = Try('journal_abbrev', ctx)
+        description_nohtml = ctx.description
+
+    def format_doi(self, doi):
+        return format_doi_as_url(self, doi)
+
+    def parse_date(self, date_str):
+        return arrow.get(dateparser.parse(date_str)).to('UTC').isoformat()
+
+class PeerJNormalizer(Normalizer):
```
chrisseto: This class is superfluous in the current state. You only have to define it if you're going to be swapping out the root parse or deserializing the data in a specific way.

providers/com/peerj/harvester.py
```diff
@@ -0,0 +1,56 @@
...
+from share import Harvester
+
+logger = logging.getLogger(__name__)
+
+
+class PeerJHarvester(Harvester):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.base_url = 'https://peerj.com/articles/index.json'
+        self.base_preprint_url = 'https://peerj.com/preprints/index.json'
```
chrisseto: It may make more sense to have a separate harvester for preprints, especially if there is the possibility of overlapping identifiers.

providers/com/peerj/harvester.py
SSJohns: Yeah, that's why I put that in there. Sometimes they will both have the number 89 or something so I just appended 'preprints-' to all the ones resulting from the preprints journal. CS already has the append in front of it.

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+class ThroughLinks(Parser):
+    link = Delegate(Link, ctx)
+
+
+class Tag(Parser):
+    name = ctx
+
+
+class ThroughTags(Parser):
+    tag = (Delegate(Tag), ctx)
```
chrisseto: `Delegate(Tag, ctx)`

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+class Publisher(Parser):
+    name = ctx
+
+
+class Institution(Parser):
+    name = RunPython('get_author_institute', ctx)
+
+    def get_author_institute(self, context):
```
chrisseto: See what happens if you run this with context being `{}`

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+    cited_name = ctx
+    order_cited = ctx('index')
+
+
+class CreativeWork(Parser):
+    title = ctx.title
+    description = Maybe(ctx, 'description-html')
+    contributors = Map(Delegate(Contributor), ctx.author)
+    links = Concat(
+        Map(Delegate(ThroughLinks), ctx.pdf_url),
```
chrisseto: `Map`'s 2nd argument is variadic.

```python
Concat(
    Delegate(ThroughLinks.using(link=Delegate(ISSN)), Maybe(ctx, 'issn')),
    Map(
        Delegate(ThroughLinks),
        ctx.pdf_url,
        RunPython('format_doi', ctx.doi),
        ctx.fulltext_html_url
    )
)
```

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+    )
+    institutions = Map(
+        Delegate(Association.using(entity=Delegate(Institution))),
+        ctx
+    )
+    subject = Delegate(Tag, ctx.subjects[0])
+    date_created = RunPython('parse_date', ctx.date)
+    date_published = RunPython('parse_date', ctx.date)
+    language = ctx.language
+    tags = Concat(
```
chrisseto: See the above comment about `Map`

`Tag` should be `ThroughTags`

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+    class Extra:
+        modified = RunPython('parse_date', ctx.date)
+        subjects = Maybe(ctx, 'subjects')
+        affiliations = Map(
+            Delegate(Association.using(entity=Delegate(Institution))),
+            ctx
+        )
+        identifiers = ctx.identifiers
+        volume = Try(ctx.volume)
+        emails = Maybe(ctx, 'author_email')
```
chrisseto: Stick with either `Try` or `Maybe`, don't mix the two.

providers/com/peerj/normalizer.py
```diff
@@ -0,0 +1,160 @@
...
+
+class Contributor(Parser):
+    person = Delegate(Person, ctx)
+    cited_name = ctx
+    order_cited = ctx('index')
+
+
+class CreativeWork(Parser):
+    title = ctx.title
+    description = Maybe(ctx, 'description-html')
```
chrisseto: Swap this with the no html description in extra. We prefer the clean title.

SSJohns: The title or the description?

providers/com/peerj/normalizer.py
SSJohns: Umm, so it auto changed context to the full dictionary in the `get_author_institution` function, but I assume you meant see what happens if `institutions` in not a value in the dictionary.

providers/com/peerj/normalizer.py
chrisseto: description

SSJohns:
@chrisseto

PR #335, opened 2016-07-29
[Feature] Elastic suggesters and multiple association fields

### LINKED ISSUE #323: Upgrade to using Elasticsearch's suggestion feature

### LINKED ISSUE #324: Index sources list using long name rather than label

### LINKED ISSUE #331: Make associations nested objects in elastic search
Want to be able to filter works by `associations.@type` and `associations.name` (or `@id`) with the guarantee that both filters match the same association. [Nested objects](https://www.elastic.co/guide/en/elasticsearch/guide/current/nested-objects.html) are meant for exactly that.

aaxelb:
Resolves #323, resolves #331, resolves #324 

## Autocompletion
Replace the `autocomplete` type with elastic's completion [suggester](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters.html).

e.g. To autocomplete a tag, POST the following to `/api/search/_suggest`:
```
{
    "tag-suggest": {
        "text": "atricle",
        "completion": {
            "field": "suggest",
            "size": 10,
            "fuzzy": true,
            "context": {
                "@type": "tag"
            }
        }
    }
}
```
`@type` can be `tag`, `person`, `institution`, `funder`, `publisher`, `organization`, or `source`.

The response will be something like:
```
{
  "_shards": {
    "total": 5,
    "successful": 5,
    "failed": 0
  },
  "tag-suggest": [
    {
      "length": 7,
      "offset": 0,
      "text": "atricle",
      "options": [
        {
          "score": 2,
          "payload": {
            "name": "Original Article",
            "@id": "15"
          },
          "text": "tag 15"
        },
        {
          "score": 2,
          "payload": {
            "name": "Article",
            "@id": "6"
          },
          "text": "tag 6"
        },
        {
          "score": 1,
          "payload": {
            "name": "Article Summary",
            "@id": "47"
          },
          "text": "tag 47"
        },
        {
          "score": 1,
          "payload": {
            "name": "Review Article",
            "@id": "64"
          },
          "text": "tag 64"
        }
      ]
    }
  ]
}
```

## Segregated associations
Instead of lumping all associations into one field, keep them separated by type for easier filtering. Works now have `institutions`, `funders`, `publishers`, and `organizations` fields.

e.g. To filter works by institution:
```
"filter": [
    {               
        "terms": {                          
            "institutions.name.raw": [                                  
                "Center For Open Science"                                                               
            ]                                                                                                                   
        }                                                                                                                                           
    }                                                                                                                                                               
]
```


bots/elasticsearch/tasks.py
```diff
@@ -37,50 +56,63 @@ def bulk_stream(self, model, ids):
...
-            'associations': [
-                self.serialize_entity(association)
-                for association in [
-                    *creative_work.funders.all(),
-                    *creative_work.publishers.all(),
-                    *creative_work.institutions.all(),
-                    *creative_work.organizations.all()
-                ]
-            ],
+            'funders': [self.serialize_entity(entity, False)
```
chrisseto: Please leave all of these on a single line, flake8 is set to ignore line length for this project.

bots/elasticsearch/tasks.py
```diff
@@ -18,6 +21,22 @@ def safe_substr(value, length=32000):
         return str(value)[:length]
     return None
 
+def add_suggest(obj):
+    input = obj['name']
+    if input:
+        # Elastic's completion suggester only looks at prefixes. If we give each
```
chrisseto: Any reason to not just use a combination of `preserve_position_increments` and `preserve_separators` set to `False`?
https://www.elastic.co/guide/en/elasticsearch/reference/current/search-suggesters-completion.html#completion-suggester-mapping

leb2dg:
Once this PR gets merged, https://github.com/CenterForOpenScience/ember-share/pull/31 should be merged as well 🎉

bots/elasticsearch/bot.py
```diff
@@ -23,51 +22,20 @@ def chunk(iterable, size):
...
-                }
-            },
-            'analyzer': {
-                'autocomplete': {
-                    'type': 'custom',
-                    'tokenizer': 'standard',
-                    'filter': ['lowercase', 'autocomplete_filter']
-                }
-            }
+SUGGEST_MAPPING = {
```
chrisseto: Could you put this inside of the ElasticSearchBot class definition?

chrisseto:
@aaxelb lgtm just a couple comments 👍

bots/elasticsearch/tasks.py
aaxelb: I hate long lines, but okay :)

bots/elasticsearch/tasks.py
aaxelb: For any combination of those two settings, the completion suggester still only matches prefixes. The example in the docs where `b` matches 'The Beatles' when `preserve_position_increments` is `false` only applies with an analyzer that removes stop words, so the analyzed string would be `beatles`.

The edge n-gram autocomplete we had before is a more natural solution to matching words in the middle of the string, but I think we'll still get the speed benefit of the completion suggester, even with a bunch of inputs per object.

chrisseto: Ah, you read closer than I do. So this will turn `'Foo Bar Baz'` into `'['Bar Baz', 'Baz']`?
Would it be better to split on whitespace? `'Foo Bar Baz'` -> `['Foo', 'Bar', 'Baz']`

bots/elasticsearch/bot.py
aaxelb: Yep yep.

bots/elasticsearch/tasks.py
aaxelb: Yeah, exactly (except with the full string `'Foo Bar Baz'` in there, too).

I think if we just split on whitespace, it would fail to match if someone types more than one word (e.g. `university of ...`).

Issue #345, opened 2016-08-03
SHARE Boolean search bug


rickjohnson:
This is a bug that appears in SHARE version 1 discovery interface, and is critical for the National Library of Medicine to use SHARE.  We have communicated it will be addressed in version 2, and this issue lays out the expectations to confirm everything is resolved within version 2 (i.e., needs to be tested and validated):

It appears that SHOULDs and MUSTs may not be applied correctly.  We need to verify the following scenarios are being applied correctly and desired results are returned in the new web discovery interface for SHARE.

From Natalie Meyers (@nkmeyers) who has been coordinating with them:

"For this query it doesn't do what I expected: 

   ?q=title:"veronese subspace codes" AND source:CrossRef

When I wast trying to test it, even If I use ?q=source:crossref or ?q=source:"CrossRef" I don't get what I'd expect then either and sometimes on browser the "data" hyperlinks are "gone" so can't click thru to see records. 

 I wonder if everything is getting interpreted as an elastic search "SHOULD" instead of "MUST" or is something else going on?  Try this one ?q=title:(veronese AND subspace) and a records with veronese in the author and NOT in the title appear in the hitlist - probably not desired behavior? "

@nkmeyers can you add more scenarios and expected results to test?

PR #351, opened 2016-08-04
ES mapping changes (don't merge until we can reindex)

### LINKED ISSUE #340: Search Queries give unexpected results
Here's the context:  I did a call yesterday with National Library of Medicine & Clinical Trials.gov Common Data Elements Task force and one of the participants asked for specific clarification about the boolean query support on SHARE webui and for me to get back to them and clarify how to do a boolean query on share web enduser interface.    

I need to show him a boolean on fields and a generic boolean.   I was going to use  this query for him as an example, but it doesn't do what I expected: ?q=title:"veronese subspace codes" AND source:CrossRef then when I wast trying to test it, even If I use ?q=source:crossref or ?q=source:"CrossRef" I don't get what I'd expect then either and sometimes on browser the "data" hyperlinks are "gone" so can't click thru to see records. 

Something weird is up with this.  I wonder if everything is getting interpreted as an elastic search "SHOULD" instead of "MUST" or is something else going on?  Try this one ?q=title:(veronese AND subspace) and a records with veronese in the author and NOT in the title appear in the hitlist - probably not desired behavior?  With whom should I consult before getting back to NLM?

### LINKED ISSUE #342: Add weight to tag _suggest

### LINKED ISSUE #343: Sort contributors by order_cited in ES

aaxelb:
### Merging will break staging until the elastic index is rebuilt!

* Sort contributors by `order_cited` (fixes #343)
* Sort suggestions based on how useful they seem (shorter, more alphabety strings are better) (fixes #342)
* More human-friendly field names for easier query composition (helps resolve #340, #345)
 * allows queries like `contributor:einstein AND title:giraffes`
 * available fields: `title`, `description`, `source`, `tag`, `language`, `date`, `award`, `venue`, `contributor`, `funder`, `publisher`, `institution`, `organization`

For local dev installs, rebuild the elastic index after pulling:
```
curl -XDELETE 'http://localhost:9200/share'
./manage.py runbot elasticsearch --last-run=0
```

### DIFF
--- a/bots/elasticsearch/tasks.py
+++ b/bots/elasticsearch/tasks.py
@@ -22,7 +22,7 @@ def safe_substr(value, length=32000):
     return None
 
 
-def add_suggest(obj):
+def add_suggest(obj, weight=None):
     if obj['name']:
         obj['suggest'] = {
             'input': re.split('[\\s,]', obj['name']) + [obj['name']],
@@ -31,7 +31,8 @@ def add_suggest(obj):
                 '@id': obj['@id'],
                 'name': obj['name'],
                 '@type': obj['@type'],
-            }
+            },
+            'weight': weight
         }
     return obj
 
@@ -78,7 +79,7 @@ class IndexModelTask(ProviderTask):
                 for affiliation in
                 person.affiliations.all()
             ],
-            'sources': [source.robot for source in person.sources.all()],
+            'sources': [self.serialize_source(source) for source in person.sources.all()],
         }
         return add_suggest(serialized_person) if suggest else serialized_person
 
@@ -128,8 +129,6 @@ class IndexModelTask(ProviderTask):
 
     def serialize_source(self, source):
         return {
-            '@id': str(source.pk),
-            '@type': 'source',
             'name': safe_substr(source.long_title),
             'short_name': safe_substr(source.robot)
         }
--- a/bots/elasticsearch/tasks.py
+++ b/bots/elasticsearch/tasks.py
@@ -124,7 +124,7 @@ class IndexModelTask(ProviderTask):
             'awards': [safe_substr(award) for award in creative_work.awards.all()],
             'venues': [safe_substr(venue) for venue in creative_work.venues.all()],
             'sources': [self.serialize_source(source) for source in creative_work.sources.all()],
-            'contributors': [self.serialize_person(person, False) for person in creative_work.contributors.all()],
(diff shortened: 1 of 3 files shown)


bots/elasticsearch/tasks.py
```diff
@@ -78,14 +86,14 @@ def serialize_person(self, person, suggest=True):
...
                 person.affiliations.all()
             ],
-            'sources': [source.robot for source in person.sources.all()],
+            'sources': [self.serialize_source(source) for source in person.sources.all()],
         }
         return add_suggest(serialized_person) if suggest else serialized_person
 
     def serialize_entity(self, entity, suggest=True):
         serialized_entity = {
-            '@id': entity.pk,
```
chrisseto: Is there any reason to remove the `@`? We were trying to make it looks a little bit like JSONLD.

bots/elasticsearch/tasks.py
```diff
@@ -94,36 +102,49 @@ def serialize_entity(self, entity, suggest=True):
...
-            'links': [safe_substr(link) for link in creative_work.links.all()],
-            'awards': [safe_substr(award) for award in creative_work.awards.all()],
-            'venues': [safe_substr(venue) for venue in creative_work.venues.all()],
-            'sources': [safe_substr(source.long_title) for source in creative_work.sources.all()],
-            'contributors': [self.serialize_person(person, False) for person in creative_work.contributors.all()],
+            'tag': [safe_substr(tag) for tag in creative_work.tags.all()],
+            'award': [safe_substr(award) for award in creative_work.awards.all()],
+            'venue': [safe_substr(venue) for venue in creative_work.venues.all()],
+            'source': [safe_substr(source.long_title) for source in creative_work.sources.all()],
+            'contributor': [c['name'] for c in serialized_lists['contributors']],
```
chrisseto: Just the names?

aaxelb: Yeah, so people can write queries like `contributor:"Chris Seto"` instead of `contributor.name`. Trying to make lucene queries in the search box a bit more human-friendly, since people apparently use and care about them. The full info for contributors and associations are included in the result payload (under `lists`), but isn't indexed.

bots/elasticsearch/tasks.py
aaxelb: This is also for human-friendliness, for queries like `type:preprint` instead of `@type:preprint`. It probably doesn't matter for `@id`; I just did that for consistency.

chrisseto: Got it! As long as we have the frontend expecting that we should be good to go.

PR #352, opened 2016-08-06
[wip] optimization for seen.add check

SSJohns:
Python will check seen.add() on every iteration to ensure it has not changed. Running it on a local variable will allow it to abstract away that change.


SSJohns:
Nvm, I think I already optimized this

PR #354, opened 2016-08-10
Feature/zenodo

SSJohns:
Zenodo updated to use the new endpoint with even more metadata for harvesting.


providers/org/zenodo/harvester.py
```diff
@@ -0,0 +1,25 @@
+import logging
+
+from furl import furl
+
+from share.harvest.oai import OAIHarvester
+
+logger = logging.getLogger(__name__)
+
+
+class ZenodoHarvester(OAIHarvester):
```
chrisseto: We don't need it just yet, but a more extendable base class for OAI might prove to be useful.

providers/org/zenodo/normalizer.py
```diff
@@ -0,0 +1,569 @@
...
+
+class Tag(Parser):
+    name = ctx
+
+
+class ThroughTags(Parser):
+    tag = Delegate(Tag, ctx)
+
+
+class CreativeWork(Parser):
```
chrisseto: Here's the fun part about zenodo, they everything from preprint to publications. We need to be able to parse that out in some way. If you happen to find a good notation for that, let me know and we can make it work.

You may be able to just make `schema` a property that just looks at `self.context` to determine what type of creative work it should be.

SSJohns: Umm we could do that in the normalizer and route each one to a different class. Then have all of those classes like preprint, publication inherit from CreativeWork or something so that they will be labeled correctly. I think that's what I did elsewhere.

providers/org/zenodo/harvester.py
SSJohns: Maybe an init with the args and redefining them? Then you'd only have to rewrite the init in each class instead of do_harvest as well?

leb2dg:
Close in favor of #372

PR #356, opened 2016-08-12
Feature/osf provider

SSJohns:
Integrating the osf preprint providers into the SHARE harvesters.


leb2dg:
Closed in favor of #360

PR #357, opened 2016-08-12
Actually use JSONLDValidator

aaxelb:
* Use `JSONLDValidator` to validate normalized data
* Update normalizers so they pass validation (don't set `date_created`)

I still get a lot of normalization errors, like missing newly required/non-null fields (e.g. `description`) and "Unresolved references" errors, where `@graph` contains `{'@id': '_:blahblah', '@type': 'blah'}` references that aren't defined elsewhere in the graph. Either we need more normalizer fixes or the validation is too strict. @chrisseto, thoughts?


share/models/validators.py
```diff
@@ -72,6 +74,9 @@ def __call__(self, value):
                 sorted(self.__blank_refs - self.__blank_nodes)
             ])))
 
+    def __eq__(self, other):
```
chrisseto: When is this used?

aaxelb: Django uses it to check whether the validator's changed and it needs to
make a migration. (Aka docs told me to
https://docs.djangoproject.com/en/1.9/topics/migrations/#adding-a-deconstruct-method
)

PR #358, opened 2016-08-15
Blank TextFields and some vaguely related fixes

aaxelb:
* Give most all `TextField`s `blank=True` (except e.g. `Tag.name` and `Person.family_name`)
* `JSONLDValidator`:
 * Require non-empty strings for non-blank text fields
 * Make it thread-safe, don't reuse reference/node sets
* Set `ExtraData.change` when created to avoid violating constraints


providers/io/osf/normalizer.py
```diff
@@ -66,7 +66,7 @@ class Project(Parser):
         ctx.embeds.affiliated_institutions.data
     )
     date_updated = tools.ParseDate(ctx.attributes.date_modified)
-    subject = tools.Delegate(Tag, ctx.attributes.category)
+    subject = tools.Delegate(Tag, tools.Maybe(ctx, 'attributes.category'))
```
chrisseto: `Try(ctx.attributes.category)`

Maybe doesn't respect `.`s 👎

share/models/change.py
```diff
@@ -242,6 +243,7 @@ def _resolve_change(self):
                 else:
                     from share.models.base import ExtraData
                     change[k] = ExtraData()
+                change[k].change = self
```
chrisseto: This seems to be handled on lines 227, 216, 184, and 165.

providers/io/osf/normalizer.py
aaxelb: I was trying to avoid creating a tag with `name=''`, which Try does when category is an empty string. It looked like Maybe worked as I wanted, but I guess it just avoided errors by never setting `subject`. How would I set `subject` only when `ctx.attributes.category` is a non-empty string?

share/models/change.py
aaxelb: Even when the `ExtraData` instance is created one line up and saved two lines down? It hits a 'change_id' non-null constraint without setting it here.

chrisseto: Ignore me, I skipped that part of elementary school where they teach you to read.

providers/io/osf/normalizer.py
chrisseto: Delete subject. It's being replaced soon and isn't currently used 🎉 
Maybe move it to tags? Map will filter out falsey values.

aaxelb: Oooh yeah, good call!

...but it doesn't seem like Map filters out falsey values. I deleted subject and made tags:
```
tags = tools.Map(tools.Delegate(ThroughTags), ctx.attributes.category, ctx.attributes.tags)
```
but that first tag still has an empty name for uncategorized projects. 😢

PR #359, opened 2016-08-15
Change DatetimeAwareJSONField to properly encode/decode things

cwisecarver:
@icereval @chrisseto This upgrades the JSONField to be compatible with datetime, date, time, and Decimal. It's fairly obvious how to add other formats. Line 22 and Line 35 show encoding and decoding, respectively. This is NOT compatible with the current format and since we're wiping the DB (and it would be very nearly impossible) I didn't write a migration. Datetimes are currently stored as epoch so.... Finding them and re-encoding them to this new format would be hard.

### DIFF
--- a/share/models/base.py
+++ b/share/models/base.py
@@ -141,7 +141,7 @@ class VersionManager(FuzzyCountManager):
 
 
 class ExtraData(models.Model, metaclass=ShareObjectMeta):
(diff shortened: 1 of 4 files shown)


cwisecarver:
Fixed a python 3 bug and added unit tests.

PR #376, opened 2016-08-19
Subjects!

### LINKED ISSUE #339: Restrict tags to PLoS's tag taxonomy
https://github.com/PLOS/plos-thesaurus

I have a json file with the parsed taxonomy available upon request. Github does not support attaching json files 👎

aaxelb:
(Replaces #374, I moved the branch to the COS repo)
* Add a `Subject` model that corresponds to a subject area in the PLOS taxonomy
 * Meant to be read-only. Trying to save a new subject will error (except bulk operations).
 * Initialize subjects with `python manage.py loaddata subjects` (added to `up.sh`)
 * If we get a new CSV from PLOS, can reload the subjects with `python manage.py reloadsubjects --max-depth=12 [subject_file] [synonym_file]`
* Replace `AbstractCreativeWork.subject` with `subjects`, a many-to-many field through `ThroughSubjects`
* Add `/api/subjects` endpoint
* Add a few tests

resolves #339


api/views/feeds.py
```diff
@@ -83,7 +83,7 @@ def item_updateddate(self, item):
         return dateparser.parse(updateddate) if updateddate else None
 
     def item_categories(self, item):
-        categories = [item.get('subject')]
+        categories = item.get('subjects')
```
chrisseto: `item.get('subjects', [])`?

share/models/change.py
```diff
@@ -251,7 +251,11 @@ def _resolve_change(self):
             elif isinstance(v, dict):
                 inst = self._resolve_ref(v)
                 change[k] = inst
-                change[k + '_version'] = inst.version
```
chrisseto: Could this be an `isinstance` check or does it result in a circular import?

tests/share/disambiguation/test_subject.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+
+from share.models import Subject
+from share.models.meta import ThroughSubjects
+from share.disambiguation import disambiguate
+
+
+def create_subject(name):
+    Subject.objects.bulk_create([
+        Subject(pk=1, name=name, lineages=[])
```
chrisseto: Do you have to specify pk here? It should auto increment.

share/models/meta.py
```diff
@@ -47,6 +49,31 @@ def __str__(self):
         return self.url
 
 
+class Subject(models.Model):
+    name = models.TextField(unique=True)
+    lineages = JSONField()
```
chrisseto: This should be a ManyToMany field, that we we get referential integrity and it will be much easier to query.

https://github.com/django-mptt/django-mptt/blob/master/mptt/fields.py#L37

api/views/feeds.py
aaxelb: I changed it to follow your suggestion, because may as well, but can many-to-many fields ever return None?

share/models/change.py
aaxelb: It could be an `isinstance` check, but I was trying to be more pythonic and duck-type it.

tests/share/disambiguation/test_subject.py
aaxelb: I don't have to specify pk, good point.

share/models/meta.py
aaxelb: Since we decided subjects could have multiple parents, `django-mptt` is out the window and getting the full lineage(s) of a subject will be slow and annoying. Since the subjects won't change except by admin intervention, it seems better not to worry about referential integrity and compute all the lineages up front.

I could add a `parents` many-to-many, so we could query for all descendents of a subject or whatever, but I feel like the `lineages` JSONField should stay to avoid reconstructing them over and over.

chrisseto:
Merged into develop some how...

PR #380, opened 2016-08-19
Fix flake8 errors

leb2dg:


tests/share/disambiguation/test_creativework.py
```diff
@@ -25,22 +26,22 @@ def test_disambiguates(self, change_id):
...
         assert disWork is None
 
     @pytest.mark.django_db
     def test_does_not_disambiguate_empty_string(self, change_id):
-        oldWork = create_work({
-            'title': '',
-            'description': 'see here is the the thing about emptiness',
-            'change_id': change_id
-        })
+        # oldWork = create_work({
```
aaxelb: For these disambiguation tests, it's important that another work exist in the database. Could you uncomment these parts and just not assign the result to `oldWork`/`oldTag` (which I assume is the problem)?

aaxelb:
You're a hero 🌿

tests/share/disambiguation/test_creativework.py
leb2dg: Yup!

project/wsgi_green.py
```diff
@@ -8,9 +8,9 @@
 """
 
 from gevent import monkey
-monkey.patch_all()
```
chrisseto: This actually has to be done in this order

project/settings.py
```diff
@@ -440,7 +442,6 @@
 HIGH_PRI_MODULES = {
 }
 
-from kombu import Queue, Exchange
```
leb2dg: @chrisseto, is this import okay to move to the top?

chrisseto: Yep, yep 👍

leb2dg:
Import fixed @chrisseto

PR #382, opened 2016-08-23
Add PeerJ and Cogprints subject mapping

leb2dg:


aaxelb:
After filling out the template, could we remove all the empty ones to make it easier to scan over the mappings that are actually defined?

providers/com/peerj/subject_mapping.json
```diff
@@ -0,0 +1,2287 @@
+{
```
aaxelb: Oh, these should be called `subject-mapping.json` with a hyphen, not underscore

PR #386, opened 2016-08-24
[Feature] docs update

leb2dg:
* update SHARE links to point to production
* update the harvester and normalizer docs
* a few small hyperlink usage changes


leb2dg:
@chrisseto, can you look over this real quick? I think the updated docs may be useful to the new interns.

PR #391, opened 2016-08-25
Change award to name on award model

leb2dg:
No normalizers were using the award field

### Changes
* change `award` field on the award model to `name`
* change field type from url --> text field


leb2dg:
@chrisseto, this should be merged before the v1 endpoint

PR #392, opened 2016-08-25
Add arxiv subject mappings

leb2dg:
### Changes
* mapping of ArXiv subjects to PLOS subjects (some map to multiple)
* update ArXiv normalizer


aaxelb:
If you change the arxiv normalizer to use the `Subjects` tool ([example](https://github.com/CenterForOpenScience/SHARE/blob/develop/providers/org/cogprints/normalizer.py#L17)), it'll map one-to-many just fine. 🐤

leb2dg:
ArXiv normalizer updated ✅

chrisseto:
@evomellor Thank you so much!

PR #394, opened 2016-08-25
Add Vagrantfile for easy setup on any machine

remram44:


remram44:
Also might I recommend using `docker-compose up worker web` in the README? This is easier to get things going than setting up a local environment.

PR #397, opened 2016-08-26
Allow multiple normalizers per user

aaxelb:
Add `app_label` field to RawData and use it to identify the normalizer to use. Allows e.g. the `io.osf` user to encompass both `io.osf` and `io.osf.preprints` documents without losing track of which normalizer to use when renormalizing.


share/migrations/0031_populate_rawdata_app_labels.py
```diff
@@ -0,0 +1,23 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.7 on 2016-08-26 15:12
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def set_app_labels(apps, schema_editor):
+    RawData = apps.get_model('share', 'RawData')
```
chrisseto: This, or something like it, will be **much** faster
```SQL
UPDATE
  rawdata
SET 
  rawdata.app_label = REPLACE(shareuser.robot, 'providers.', '')
JOIN
  shareuser
ON
  rawdata.source = shareuser.id
```

pinging @cwisecarver for second thoughts

aaxelb: That looks so much better.

cwisecarver: This works. It does two sequence scans but I don't think there's a way around it since the replace has to get the value of every row. It's better than looping through it in python knowing what we know about the django queryset. I would say however, best to do this after hours since I imagine it will totally lock the table it's writing to.

```sql
UPDATE
  share_rawdata
SET
  app_label = REPLACE(share_shareuser.robot, 'providers.', '')
FROM
  share_shareuser
WHERE share_rawdata.source_id = share_shareuser.id;

cwisecarver: @aaxelb @chrisseto ^^

PR #403, opened 2016-08-29
Temporary Preprint Harvesters from OSF For Meetings Projects

erinspace:
Make some temporary changes to the current OSF related preprint harvesters that were ready to go - harvest from the respective tags and get the current projects stored in OSF For Meetings as a starter until preprint branding is ready

- Add psyarxiv
- Change URL to projects url filtering on tags
- Add separate normalizers just using their own Subject as a static element (sorry for all the extra stuff, it was done in the name of speed... tried a few other options and could have gotten something way better with a bit more time...)
- add an embeds url


providers/io/socarxiv/normalizer.py
```diff
@@ -0,0 +1,112 @@
...
+        tools.Delegate(Association.using(entity=tools.Delegate(Institution))),
+        ctx.embeds.affiliated_institutions.data
+    )
+    date_updated = tools.ParseDate(ctx.attributes.date_modified)
+    tags = tools.Map(tools.Delegate(ThroughTags), ctx.attributes.category, ctx.attributes.tags)
+    rights = tools.Maybe(ctx, 'attributes.node_license')
+    links = tools.Map(tools.Delegate(ThroughLinks), ctx.links.html)
+    subjects = tools.Map(
+        tools.Delegate(ThroughSubjects),
+        tools.Concat(tools.Static({'text':'Social and behavioral sciences'}))
```
leb2dg: Why is this a dictionary?

erinspace: I got a `keyError` when running - it was looking for a text element! So this made things work, but I kinda hacked it so perhaps am doing it wrong?

providers/io/engrxiv/harvester.py
```diff
@@ -16,13 +16,16 @@ class EngrxivHarvester(OSFHarvester):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.url = '{}v2/preprint_providers/engrxiv/preprints/'.format(settings.OSF_API_URL)
+        # self.url = '{}v2/preprint_providers/engrxiv/preprints/'.format(settings.OSF_API_URL)
+        # temporary harvest URL until preprint branding is complete
+        self.url = '{}v2/nodes/?filter[tags]=engrxiv'.format(settings.OSF_API_URL)
```
aaxelb: OSFHarvester has `self.url = 'https://api.osf.io/v2/nodes/'`, so instead of changing it in `__init__` you could just add `url.args['filter[tags]'] = 'engrxiv'` in do_harvest down below. (It was using the staging api in my local testing, which had 0 results)

providers/io/socarxiv/normalizer.py
aaxelb: The `Subject` parser a few lines up has `name = ctx.text`, is the only reason it's looking for 'text'. If you change that to `name = ctx` it'll accept a plain string here.

leb2dg: Oh weird. I did a similar thing for [bioRxiv](https://github.com/CenterForOpenScience/SHARE/blob/develop/providers/org/biorxiv/normalizer.py#L61) but didn't use a dictionary. Then in the subject class it was just `ctx` instead of `ctx.text`. But as long as it works it doesn't really matter.

aaxelb:
Ran into a weird thing with this: When harvesting preprints through `/v2/nodes/`, if the node had already been harvested as an OSF project, it would be disambiguated to the existing project (even though `preprint` and `project` are different types?) with no changes except `date_updated` and a new `Subject`. The changeset would still have a reference with `'@type': 'preprint'`, causing an error when trying to accept the changes. 

Need to look at why the disambiguator matches a project for a preprint (does it actually use `abstractcreativework`?) but as a temporary fix, we decided to append " -- Preprint" to the titles of `*rxiv.io` preprints. It prevents them being disambiguated and they show up properly as preprints.

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    url = ctx
+    type = tools.Static('provider')
+
+
+class ThroughLinks(Parser):
+    link = tools.Delegate(Link, ctx)
+
+
+class Subject(Parser):
+    name = ctx.text
```
aaxelb: This should just be `ctx`

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+        locale = ctx.embeds.users.data.attributes.locale
+        date_registered = ctx.embeds.users.data.attributes.date_registered
+        active = ctx.embeds.users.data.attributes.active
+        timezone = ctx.embeds.users.data.attributes.timezone
+        profile_image = ctx.embeds.users.data.links.profile_image
+
+
+class Contributor(Parser):
+    person = tools.Delegate(Person, ctx)
+    order_cited = ctx('index')
```
chrisseto: OSF sends back indexes https://github.com/CenterForOpenScience/SHARE/blob/develop/providers/io/osf/preprints/normalizer.py#L57

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
+from share.normalize import ctx
+from share.normalize.parsers import Parser
+from share.normalize import tools
+
+
+class Identifier(Parser):
+    url = ctx
+    base_url = tools.Static('https://osf.io/')
```
chrisseto: Making this dynamic will make it easier to test on staging https://github.com/CenterForOpenScience/SHARE/blob/develop/providers/io/osf/preprints/normalizer.py#L12

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    url = ctx.embeds.users.data.links.html
+    identifiers = tools.Map(tools.Delegate(ThroughIdentifiers), ctx.embeds.users.data.links.html)
+
+    class Extra:
+        nodes = ctx.embeds.users.data.relationships.nodes.links.related.href
+        locale = ctx.embeds.users.data.attributes.locale
+        date_registered = ctx.embeds.users.data.attributes.date_registered
+        active = ctx.embeds.users.data.attributes.active
+        timezone = ctx.embeds.users.data.attributes.timezone
+        profile_image = ctx.embeds.users.data.links.profile_image
```
chrisseto: This should be an identifier

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    name = ctx.text
+
+
+class ThroughSubjects(Parser):
+    subject = tools.Delegate(Subject, ctx)
+
+
+class Preprint(Parser):
+
+    def update_title_for_preprint(self, title):
```
chrisseto: Is this necessary?

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+        tools.Delegate(Association.using(entity=tools.Delegate(Institution))),
+        ctx.embeds.affiliated_institutions.data
+    )
+    date_updated = tools.ParseDate(ctx.attributes.date_modified)
+    tags = tools.Map(tools.Delegate(ThroughTags), ctx.attributes.category, ctx.attributes.tags)
+    rights = tools.Maybe(ctx, 'attributes.node_license')
+    links = tools.Map(tools.Delegate(ThroughLinks), ctx.links.html)
+    subjects = tools.Map(
+        tools.Delegate(ThroughSubjects),
+        tools.Concat(tools.Static('Engineering and technology'))
```
chrisseto: Shouldn't need the concat here

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    title = title = tools.RunPython('update_title_for_preprint', ctx.attributes.title)
+    description = ctx.attributes.description
+    contributors = tools.Map(tools.Delegate(Contributor), ctx['contributors'])
+    institutions = tools.Map(
+        tools.Delegate(Association.using(entity=tools.Delegate(Institution))),
+        ctx.embeds.affiliated_institutions.data
+    )
+    date_updated = tools.ParseDate(ctx.attributes.date_modified)
+    tags = tools.Map(tools.Delegate(ThroughTags), ctx.attributes.category, ctx.attributes.tags)
+    rights = tools.Maybe(ctx, 'attributes.node_license')
```
chrisseto: Maybe doesn't support `.`s Use `Try(ctx.attributes.node_license)`

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+    links = tools.Map(tools.Delegate(ThroughLinks), ctx.links.html)
+    subjects = tools.Map(
+        tools.Delegate(ThroughSubjects),
+        tools.Concat(tools.Static('Engineering and technology'))
+    )
+
+    class Extra:
+        date_created = tools.ParseDate(ctx.attributes.date_created)
+        files = ctx.relationships.files.links.related.href
+        parent = tools.Maybe(ctx, 'relationships.parent.links.related.href')
```
chrisseto: See above note about maybe

providers/io/engrxiv/normalizer.py
aaxelb: It was until I merged #408 just now

providers/io/engrxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+        registrations = ctx.relationships.registrations.links.related.href
+        logs = ctx.relationships.logs.links.related.href
+        node_links = ctx.relationships.node_links.links.related.href
+        wikis = ctx.relationships.wikis.links.related.href
+        children = ctx.relationships.children.links.related.href
+        fork = ctx.attributes.fork
+        date_modified = ctx.attributes.date_modified
+        collection = ctx.attributes.collection
+        registration = ctx.attributes.registration
+        type = ctx.type
```
chrisseto: This will always be `'nodes'` we can probably ignore it.

providers/io/psyarxiv/normalizer.py
```diff
@@ -0,0 +1,115 @@
+from share.normalize import ctx
```
chrisseto: Could you just point all the other providers to the same normalizer?

providers/io/socarxiv/normalizer.py
```diff
@@ -0,0 +1,116 @@
...
+        tools.Delegate(Association.using(entity=tools.Delegate(Institution))),
+        ctx.embeds.affiliated_institutions.data
+    )
+    date_updated = tools.ParseDate(ctx.attributes.date_modified)
+    tags = tools.Map(tools.Delegate(ThroughTags), ctx.attributes.category, ctx.attributes.tags)
+    rights = tools.Maybe(ctx, 'attributes.node_license')
+    links = tools.Map(tools.Delegate(ThroughLinks), ctx.links.html)
+    subjects = tools.Map(
+        tools.Delegate(ThroughSubjects),
+        tools.Concat(tools.Static('Social and behavioral sciences'))
```
chrisseto: Related to previous comment, you can use subclassing to handle different tags on everything. I think cogprints is the best example of this?

PR #409, opened 2016-08-30
v1 Push api endpoint and normalizer

leb2dg:
* endpoint only accepts one submission at a time


api/views/workflow.py
```diff
@@ -168,3 +170,96 @@ def get_queryset(self):
...
+    permission_classes = [ReadOnlyOrTokenHasScopeOrIsAuthenticated, ]
+    serializer_class = NormalizedDataSerializer
+
+    def get_queryset(self):
+        return NormalizedData.objects.all()
+
+    def post(self, request, *args, **kwargs):
+
+        prelim_data = request.data
+        prelim_data['source'] = ShareUser.objects.get(id=request.user.id)
```
aaxelb: I don't think you have to put the user in `prelim_data`, you can just pass it to `store_data` separately.

leb2dg: Removing source from `prelim_data` gets rid of the serialize error but source is a required field.
```
"errors": {
        "source": [
            "This field is required."
        ],
        "provider_doc_id": [
            "This field is required."
        ],
        "sha256": [
            "This field is required."
        ],
        "data": [
            "This field is required."
        ],
        "app_label": [
            "This field is required."
        ]
    },
```

leb2dg: Hmm, nevermind. Seems like `store_data` isn't doing it's job.

share/tasks.py
```diff
@@ -136,6 +136,7 @@ def run(self, normalized_id, started_by_id=None):
             cs = ChangeSet.objects.from_graph(ChangeGraph.from_jsonld(normalized.normalized_data, extra_namespace=normalized.source.username), normalized.id)
             if cs and normalized.source.is_robot:
                 # TODO: verify change set is not overwriting user created object
+                # TODO: add flag for other sources that aren't robots
```
leb2dg: @chrisseto, do you have any preferences/ideas on how this should be done?

leb2dg:
@chrisseto, this is ready for review 🏦  Also, should the original v1 schema be included anywhere? Right now it is just at the top level.

api/views/workflow.py
```diff
@@ -10,8 +10,12 @@
 from share.models import ChangeSet, Change, RawData, ShareUser, NormalizedData
 from share.models.validators import JSONLDValidator
 from share.tasks import MakeJsonPatches
+from share.harvest.harvester import Harvester
+from share.normalize.v1_push import V1Normalizer
+from django.db import transaction
```
chrisseto: Imports should be ordered by stdlib, 3rd part, application and separated by a single blank line.

```python
import os  # stdlib

from django import thing  # 3rd party

from share.whatever import bar  # application
```

api/views/workflow.py
```diff
@@ -234,3 +238,96 @@ def get(self, request, *args, **kwargs):
...
+
+    def post(self, request, *args, **kwargs):
+
+        prelim_data = request.data
+        app_label = request.user.username
+
+        # store raw data, assuming you can only submit one at a time
+        raw = None
+        with transaction.atomic():
+            doc_id = prelim_data['uris']['canonicalUri']
```
chrisseto: You'll need a try catch here or some form of validation to ensure you won't get a `KeyError`

api/views/workflow.py
```diff
@@ -234,3 +238,96 @@ def get(self, request, *args, **kwargs):
...
+            raw = RawData.objects.store_data(doc_id, Harvester.encode_json(self, prelim_data), ShareUser.objects.get(id=request.user.id), app_label)
```
chrisseto: You should be able to just pass `request.user` here.

share/migrations/0039_update_trigger_migrations_20160906_1801.py
```diff
@@ -0,0 +1,184 @@
+# -*- coding: utf-8 -*-
```
chrisseto: You don't need to add a trigger update unless you change something that inherits from `ShareObject`

share/normalize/v1_push.py
```diff
@@ -0,0 +1,338 @@
...
+    rights = tools.Join(tools.Try(ctx.licenses.uri))
+
+    subjects = tools.Map(
+        tools.Delegate(ThroughSubjects),
+        tools.Try(ctx.subjects)
+    )
+
+    tags = tools.Map(
+        tools.Delegate(ThroughTags),
+        tools.Concat(
```
chrisseto: You shouldn't need the `Concat` here, it's implied in `Map`

share/normalize/v1_push.py
```diff
@@ -0,0 +1,338 @@
...
+
+    # Note: this is only taking the first language in the case of multiple languages
+    language = tools.ParseLanguage(
+        tools.Try(ctx.languages[0]),
+    )
+
+    links = tools.Concat(
+        tools.Map(
+            tools.Delegate(ThroughLinks),
+            tools.Concat(
```
chrisseto: See below comment about `Map`

leb2dg:
@chrisseto, CR addressed 📦

api/urls.py
```diff
@@ -40,6 +40,7 @@
     url(r'rss/?', views.CreativeWorksRSS(), name='rss'),
     url(r'atom/?', views.CreativeWorksAtom(), name='atom'),
     url(r'userinfo/?', views.ShareUserView.as_view(), name='userinfo'),
+    url(r'v1data/?', views.V1DataView.as_view(), name='v1data'),
```
chrisseto: Could we map this to the [existing url}(https://osf.io/api/v1/share/data/help/#!/SHARE/post_share_data)?
I think it is `/api/v1/share/data` or something similar.

leb2dg: Do you mean change the name so it's the same as the old one? Or would mapping the new to the old involve something more?

share/normalize/v1_push.py
```diff
@@ -0,0 +1,337 @@
...
+    links = tools.Concat(
+        tools.Map(
+            tools.Delegate(ThroughLinks),
+            tools.Try(ctx.uris.canonicalUri),
+            tools.Try(ctx.uris.descriptorUris),
+            tools.Try(ctx.uris.objectUris)
+        ),
+        tools.Map(
+            tools.Delegate(ProviderThroughLinks),
+            tools.Concat(
```
chrisseto: `Concat` is not necessary here

api/views/workflow.py
```diff
@@ -234,3 +238,102 @@ def get(self, request, *args, **kwargs):
...
+        raw = None
+        with transaction.atomic():
+            try:
+                doc_id = prelim_data['uris']['canonicalUri']
+            except KeyError:
+                return Response({'errors': 'Canonical URI is required.', 'data': prelim_data}, status=status.HTTP_400_BAD_REQUEST)
+            except:
+                return Response({'errors': 'An error was encountered. Please ensure the data is formatted properly.', 'data': prelim_data}, status=status.HTTP_400_BAD_REQUEST)
+
+            raw = RawData.objects.store_data(doc_id, Harvester.encode_json(self, prelim_data), request.user, app_label)
```
leb2dg: @chrisseto, this is where the raw data is being stored

api/views/workflow.py
```diff
@@ -234,3 +238,102 @@ def get(self, request, *args, **kwargs):
...
+        app_label = request.user.username
+
+        # store raw data, assuming you can only submit one at a time
+        raw = None
+        with transaction.atomic():
+            try:
+                doc_id = prelim_data['uris']['canonicalUri']
+            except KeyError:
+                return Response({'errors': 'Canonical URI is required.', 'data': prelim_data}, status=status.HTTP_400_BAD_REQUEST)
+            except:
```
chrisseto: If it is not an expected error we should just let django handle it and send back a 500 to the user.

v1_push_api.json
```diff
@@ -0,0 +1,348 @@
+{
```
chrisseto: Is this being used anywhere?

api/views/workflow.py
```diff
@@ -234,3 +238,102 @@ def get(self, request, *args, **kwargs):
...
+                            }
+                        }
+
+                    }
+        Success:       200 OK
+    """
+    permission_classes = [ReadOnlyOrTokenHasScopeOrIsAuthenticated, ]
+    serializer_class = NormalizedDataSerializer
+
+    def get_queryset(self):
```
chrisseto: I don't think it would be appropriate to support `GET`ing on this endpoint as there is already another endpoint that returns this data.

v1_push_api.json
leb2dg: Nope, I was just using it as a reference and you never gave your opinion on this:
>
Also, should the original v1 schema be included anywhere? Right now it is just at the top level.
>

chrisseto: Ah, at the least it should be used for data validation in the endpoint. I don't see a reason to not expose it otherwise.

leb2dg:
@chrisseto, comments addressed 📼

chrisseto:
@leb2dg Looks like you have some flake8 issues and migrations issues. Merge/rebase develop and move your migration to the "top"

PR #410, opened 2016-08-30
URI format validation

aaxelb:
* Remove the ISSN non-url Link from the PeerJ normalizer (every preprint has a Link with `url='2167-9843'`, which could stand to be deleted). Put issn in extra instead.
 * Do the same for com.springer and com.biomedcentral
* Make sure that sort of thing never happens again by putting `'format': 'uri'` in the `JSONLDValidator` schema for `ShareURLField`s
 * Also make `Link.url` a `ShareURLField`, fixing an accidental oversight?
 * Test the format validation


share/models/validators.py
```diff
@@ -3,29 +3,15 @@
...
 from django.db import connection
 from django.apps import apps
 from django.core.exceptions import ValidationError
 from django.utils.deconstruct import deconstructible
 
-
-def is_valid_uri(value):
-    # uri = rfc3987.get_compiled_pattern('^%(URI)s$')
-    try:
-        assert regex.match(r'.+//:[^/]/+.*', value)
```
chrisseto: Pinging @cwisecarver to see if he remembers what weird data we were trying push through this.

cwisecarver: I think we wanted it to be able to match more esoteric URIs that don't look like `http://something.com/stuff`. IIRC stuff from JSON-LD and schema.org.

aaxelb: The `draft4_format_checker` uses the `rfc3987` module with format `uri`, which seems pretty permissive. It matches URIs like `urn:ISSN:1234-1234`. Are there technically invalid URIs we want to accept?

chrisseto: @aaxelb I think that's the strangest thing we wanted to match. I don't think anything properly formats ISSNs currently, we'll have to keep watch on staging for errors 👍

chrisseto:
@aaxelb Holding off on merging this for a little while as it will bring a slew of normalizer updates with it. There are a non-trivial amount of incorrectly formatted DOIs floating around. (I think)

chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/commit/e430f4cdea767b380d4df64ff0c3f0a0f07acd14

PR #415, opened 2016-09-06
[SHARE-427] Feature/philica

zamattiac:


leb2dg:
There are some flake8 errors in the Normalizer file

chrisseto:
Preface: *Potentially a dumb idea*

Phillica doesn't look like they support pagination, but we do know the date of the first publication,  23rd April, 2006 at 16:50:12. We could do a binary search on their API to actually get the full date range requested by the harvest.

zamattiac:
Flake should be good now, even though I still get *invalid syntax* on line 18 char 37... a blank line.

PR #418, opened 2016-09-07
Only label RePEc's preprints as preprints

### LINKED ISSUE #417: Update RePec normalizer so not everything is categorized as a Preprint
* determine if this is possible based on their data

leb2dg:
Resolves #417 

* this also impacts any OAI creativework that has "preprint" as the type

### DIFF
--- a/share/normalize/oai.py
+++ b/share/normalize/oai.py
@@ -116,7 +116,7 @@ class OAICreativeWork(Parser):
             if resourceType == 'preprint':
                 return 'Preprint'
             return 'CreativeWork'
-        except KeyError:
+        except (KeyError, ValueError):
             return 'CreativeWork'
 
     ORGANIZATION_KEYWORDS = (
--- a/share/normalize/oai.py
+++ b/share/normalize/oai.py
@@ -109,7 +109,15 @@ class OAIThroughTags(Parser):
 
 
 class OAICreativeWork(Parser):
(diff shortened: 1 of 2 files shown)


aaxelb:
Are there other `dc:type` strings that are commonly used? It'd be awesome if we could recognize Publications and minimize our use of CreativeWork as a catch-all

leb2dg:
`type` isn't a controlled field for OAI but some sources use consistent types so there is some room for improvement.

share/normalize/oai.py
```diff
@@ -109,7 +109,15 @@ class OAIThroughTags(Parser):
...
 class OAICreativeWork(Parser):
-    schema = 'CreativeWork'
+    @property
+    def schema(self):
+        try:
+            resourceType = self.context['record']['metadata']['dc']['dc:type']
+            if resourceType == 'preprint':
+                return 'Preprint'
+            return 'CreativeWork'
+        except KeyError:
```
chrisseto: You may want to catch `ValueError` here as well just in case one of the key returns `None`

leb2dg:
@chrisseto, comment addressed 🚝

chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/commit/6ad28cf3ebb3f1ff7f95603f5f9312a25bd9c376

PR #421, opened 2016-09-09
Update urls to v2

leb2dg:
Should be merged in conjunction with: https://github.com/CenterForOpenScience/ember-share/pull/110

* Updated url on OSF


chrisseto:
Couldn't you just change the root url in `project/urls`?

```python
    url(r'^api/', include('api.urls', namespace='api')),
```

leb2dg:
Ahh I figured there was a better way. Will change 👍

leb2dg:
@chrisseto, the API changes are ready for review/testing 💸

chrisseto:
And throwing a deployment note here. Preprints relies on the API without the v2. We'll have to coordinate releasing the two.

chrisseto:
Actually, it might make more sense to just throw a redirect in the django app. That way we can just redirect `/api/(?!v\d+)/` to the latest api version.

leb2dg:
Will that work with differentiating between v1 and v2 urls?

chrisseto:
It should. Just match all api urls that don't have `vsomething` directly after the `/api/` bit

leb2dg:
Oh nice, I'll add that then

PR #427, opened 2016-09-16
[Feature] Update v1 normalizer tests

leb2dg:
* Rename test file
* Check that changesets were accepted

### DIFF
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -336,7 +336,6 @@ class V1DataView(views.APIView):
         # normalize data
         normalized_data = V1Normalizer({}).normalize(raw.data)
         data = {}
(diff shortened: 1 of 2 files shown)


chrisseto:
@leb2dg Could you update this when you have a moment?

leb2dg:
@chrisseto, resolved conflicts

leb2dg:
Tests are now broken though...will fix

api/views/workflow.py
```diff
@@ -338,7 +338,7 @@ def post(self, request, *args, **kwargs):
         data = {}
         data['source'] = request.user.id
```
chrisseto: Could you remove this line? It's handled by passing request in (or at least should be)

Issue #404, opened 2016-08-30
Link Related Works in SHARE


rickjohnson:
The SHARE data model current supports linking different related entities (e.g. researcher to institution, researcher to publication, etc).  However, linking items of the same entity type is not supported yet (esp. dataset to publication, dmp to dataset, dmp to osf project).  Any abstractcreativework should be able to be linked to another abstractcreativework.  While I assume these links will be mapped via the RDBMS schema, they would ideally implement the patterns in RMAP for asserting relationships:  http://rmap-project.info/rmap/wp-content/uploads/JCDL2015_Poster_HandoutSize.pdf (see labels on arrows under "How it Works")

https://osf.io/j34su/.  (see slide 2 of that presentation showing the relationship patterns).  SHARE supports all of this now EXCEPT the work to work assertions.)

Changes such as the following are likely needed:
- Add join table for two related works that includes two abstractcreativework ids and field for relationship type (maybe using RMAP Disco: http://rmap-project.info/rmap/wp-content/uploads/JCDL2015_Poster_HandoutSize.pdf).  Also, ideally it uses the URI value for the relationship.  Will see if I can track that down.
- Likely changes to harvester, push API, and curation tools to assert these relationships
- Likely changes to ember-share app to filter on relationships (i.e. new filter tag on left sidebar)
- Opportunities for visualizing relationships of objects in a graph

rickjohnson:
I assume this should have label 'Epic', but I do not see how to add that label (I may not have permission)

rickjohnson:
Here is link to more info about RMAP:

https://rmap-project.atlassian.net/wiki/display/RMAPPS/Data+Model

rickjohnson:
Thinking more table could just mirror subject object predicate where subject and object are primary keys for objects in abstractcreativework and predicate is a machine addressable URI within a community ontology

This may be most comprehensive ontology to use first:

http://www.sparontologies.net/ontologies/cito/source.html#d4e101

Under object properties examples include:

uses_data_from, uses_method_in, describes, retracts, cites, derides, is_refuted_by

So example row could be
subject object predicate
12356   45678 uses_data_from

Where 12356 is an article and 45678 is a dataset (in a sentence would be: article 12356 uses data from dataset 45678)

laurenbarker:
Closing since SHARE supports work to work relations now.

PR #429, opened 2016-09-22
[SHARE-361] [Feature] Work relations and identifiers

aaxelb:
* Relations!
 * Add `Relation`, a through table from creative work to creative work
 * Allowed relation types in `share/models/relation-types.json`
* Identifiers!
 * Rename `Identifier` to `PersonIdentifier`
 * Add `CreativeWorkIdentifier`
 * Both types of identifiers are many-to-one to the things they identify
 * Use identifiers for disambiguating people and works
  * Allow changing a work's type by submitting a changeset for a work with matching identifier and different `@type` (unless the new type is `creativework`)
* Remove `Link` model. The links harvested with a work should be identifiers either for that work or for a related work.
* Update some normalizers to use identifiers/relations instead of links
 * io.osf, org.crossref work
 * The rest of the normalizers can wait until we've settled other pending data model changes


share/models/creative.py
```diff
@@ -12,18 +12,22 @@
...
+    identifiers = ShareManyToManyField('Identifier', through='WorkIdentifier')
 
     # Used to determine if something should be surfaced in ES or not
     # this may need to be renamed later
     is_deleted = models.BooleanField(default=False)
 
+    # True for works that only exist because they were referenced by other works.
+    # Should be false if there's any information beyond an identifier.
+    # Don't put stubs in ES.
+    is_stub = models.BooleanField(default=False)
```
chrisseto: I'm don't think this field is necessary. We can just define a minimum set of fields required to be indexed into elastic if this does become a problem. I just foresee a lot of issues involving either flipping this by mistake or forgetting to set it to false.

I see why this would be nice to have in the database but it's a field that is dependant upon the state of other fields. Any case like that I'd prefer to use `@property` to define them and forfeit the ability to filter it at the database level., that way it's impossible to be out of sync.

Side note: While it's a super edge case, I could see the use case for finding mostly empty records in elasticsearch to fill them out based on where the identifier resolve to.

share/models/change.py
```diff
@@ -177,6 +177,8 @@ def _create(self, save=True):
             except IntegrityError as e:
                 from share.disambiguation import disambiguate
                 logger.info('Handling unique violation error %r', e)
+                if self.target_type.model == 'identifier':
```
chrisseto: Having a unique violation on identifier should be an acceptable case. It is often unavoidable when handling large amounts of data. Anytime a `MakeJsonPatches` job is run while there are pending changes this block could be hit.

share/disambiguation.py
```diff
@@ -145,19 +144,25 @@ class AbstractCreativeWorkDisambiguator(Disambiguator):
...
-        if self.attrs.get('links'):
-            for link in self.attrs.get('links'):
-                models = ThroughLinks.objects.select_related('creative_work', 'link').filter(link=link)[:2].all()
-                if len(models) == 1 and 'issn' not in models[0].link.type.lower():
-                    return models[0].creative_work
+        if self.attrs.get('identifiers'):
+            for id in self.attrs.get('identifiers'):
+                try:
+                    identifier = Identifier.objects.get(id=id)
+                    return identifier.abstractcreativework_set.first()
```
chrisseto: `.first()` is **very** slow, avoid it at all costs.

share/models/people.py
```diff
@@ -16,25 +16,17 @@ def __str__(self):
         return self.email
 
 
-class Identifier(ShareObject):
```
chrisseto: I'm not certain that combining `Link`s and `Identifier`s is the right thing to do. It makes sense from a simplicity standpoint. I'd be curious to see the performance of  splitting heavy reading/writing between two tables or just one. `Identifier` will be the most active table in the database, it probably is already.

share/models/meta.py
```diff
@@ -1,17 +1,41 @@
...
+    def get_base_url(self):
+        return furl(self.url).origin
+
+
+class Relation(ShareObject):
+    with open('./share/models/relation-types.json') as fobj:
+        # TODO add label to file
+        RELATION_TYPES = [(t['key'], t['uri']) for t in json.load(fobj)]
+
+    subject_work = ShareForeignKey('AbstractCreativeWork', related_name='outgoing_%(class)ss')
```
aaxelb: I named them after the "subject-object-predicate" approach Rick [suggested](https://github.com/CenterForOpenScience/SHARE/issues/404#issuecomment-243841316), but you're right, `from` and `to` are probable clearer, considering how overloaded "subject" and "object" already are.

share/models/meta.py
```diff
@@ -1,17 +1,41 @@
...
+    object_work = ShareForeignKey('AbstractCreativeWork', related_name='incoming_%(class)ss')
+    relation_type = models.TextField(choices=RELATION_TYPES, blank=True)
```
aaxelb: Yes, for when a work has a link to something but we don't know how it's related. Would it be better to have an explicit `unknown` type, do you think?

share/models/creative.py
aaxelb: That makes perfect sense, I'll remove it.

I mostly added `is_stub` not for ES, but to distinguish between an empty change that means "no changes, skip" and an empty change that means "create an empty object", but I can do that without an actual field.

share/models/change.py
aaxelb: I agree it's unavoidable, but the way we handle it will always result in an identifier pointing to multiple objects, and often at least one relation pointing to a "stub" instead of the work it actually relates to. That doesn't seem acceptable, and rewinding a few `MakeJsonPatches` tasks seemed better than trying to re-disambiguate the identified work/person at this stage. 

I could resolve the race condition without the task actually failing, though, by catching this particular exception in `MakeJsonPatches`, so it can retry more gracefully...

share/disambiguation.py
aaxelb: Duly noted.

share/models/people.py
aaxelb: I've gone back and forth on this, and went with a single table mostly because it was easier to name (`Link` sounds more like a relation than an identifier, and `PersonIdentifier`/`WorkIdentifier` seemed cumbersome...). Separate identifier tables will almost certainly have better performance, I think, so I'll split them up.

chrisseto: Just chatted with the local SQL expert @cwisecarver. The table's indexes should always be able to fit in memory so it should be fine to put everything in one 👍

share/models/creative.py
chrisseto: We can just make it so empty create requests are not skipped

share/models/meta.py
chrisseto: I think so? It feels more correct to stay within `choices` whenever they are specified.

share/models/meta.py
```diff
@@ -1,17 +1,41 @@
...
 
 # TODO Rename this file
 
 
+class Identifier(ShareObject):
+    # https://twitter.com/berniethoughts/
+    url = ShareURLField(unique=True)
+
+    # https://twitter.com/
+    def get_base_url(self):
```
chrisseto: I think there will be some benefit to storing this or something like type. That way we'll be able to easily get a count of how many DOIs there are or how many people have OSF profiles. 

Would `domain` make more sense?

tests/share/changes/test_change.py
```diff
@@ -193,43 +195,42 @@ def test_can_delete_work(self, john_doe, normalized_data_id):
...
                 '@type': 'preprint',
                 'title': 'All About Cats',
+                'identifiers': [{'@id': '_:def', '@type': 'workidentifier'}]
             }]
         })
 
         change_set = ChangeSet.objects.from_graph(graph, normalized_data_id)
 
-        link, preprint, _ = change_set.accept()
+        preprint = [o for o in change_set.accept() if isinstance(o, Preprint)][0]
```
chrisseto: This is a subtle but very important nuance. The order that these are returned in is critically import to the disambiguation process.

There are a few other tests that check this but I'd sleep better know that there are many assertions for it.

See `test_topological_sort_many_to_many` for another example.

share/models/validators.py
```diff
@@ -118,7 +118,7 @@ def json_schema_for_field(self, field):
             return rel
         if field.choices:
             return {
-                'enum': field.choices,
+                'enum': [c[0] for c in field.choices],
```
chrisseto: Won't this break choices fields that are just a list of strings?

share/models/meta.py
```diff
@@ -1,17 +1,41 @@
...
+        return furl(self.url).origin
+
+
+class Relation(ShareObject):
+    with open('./share/models/relation-types.json') as fobj:
+        # TODO add label to file
+        RELATION_TYPES = [(t['key'], t['uri']) for t in json.load(fobj)]
```
chrisseto: Is there any benefit to storing the uri? From the looks of it the uri's can be computed if need be?

share/models/change.py
```diff
@@ -189,6 +191,12 @@ def _create(self, save=True):
 
     def _update(self, save=True):
         self.target.change = self
+
+        new_type = self.change.pop('@type', None)
+        if new_type:
+            # TODO is this enough?
```
chrisseto: I think there is a method on TypedObjects called `transition_type` or `recast_as` that will handle the conversion for you.

share/disambiguation.py
chrisseto: Shouldn't identifiers only ever have a single creativework pointing towards them?

share/disambiguation.py
```diff
@@ -121,6 +112,14 @@ class PersonDisambiguator(Disambiguator):
     FOR_MODEL = Person
 
     def disambiguate(self):
+        if self.attrs.get('identifiers'):
+            for id in self.attrs.get('identifiers'):
+                try:
+                    identifier = Identifier.objects.get(id=id)
+                    return identifier.person_set.first()
```
chrisseto: 🔥  `.first()` 🔥

share/disambiguation.py
```diff
@@ -2,16 +2,7 @@
...
-from share.models import Link
-from share.models import Person
-from share.models import Subject
-from share.models import Contributor
-from share.models import Association
-from share.models import Affiliation
-from share.models import PersonEmail
-from share.models.meta import ThroughLinks
-from share.models import AbstractCreativeWork
+from share.models import Tag, Person, Subject, Contributor, Association, Affiliation, PersonEmail, AbstractCreativeWork, Identifier, Relation
```
chrisseto: Probably better to just `from share import models` at this point 🍤

share/change.py
```diff
@@ -55,14 +56,28 @@ def refs(self):
         return self.__refs
 
     @property
+    def new_type(self):
+        return self.__new_type or self.type
+
+    @new_type.setter
+    def new_type(self, value):
+        # CreativeWork is the catch-all/unknown type. Don't overwrite a more specific type
+        self.__new_type = value if value != 'creativework' else None
```
chrisseto: This check should be done at the acceptance or disambiguation stage

share/change.py
```diff
@@ -193,7 +218,12 @@ def __init__(self, graph):
         self.__graph = graph
         self.__visted = set()
         self.__visiting = set()
-        self.__nodes = list(graph.nodes)
+
+        # visit identifiers first
+        self.__nodes = sorted(
```
chrisseto: This should be handled properly in the existing logic. If it is not, resorting the entire list is not the correct way to fix it.

share/change.py
```diff
@@ -108,12 +130,15 @@ def __init__(self, node, disambiguate=True, extra_namespace=None):
...
             self._disambiguate()
 
-    def resolve_relations(self, mapper):
+    def resolve_through_relations(self, mapper):
         for key, relations in self.__reverse_relations.items():
-            resolved = [
-                next(x for x in mapper[r['@id'], r['@type'].lower()].related if x['@id'] != self.id)
-                for r in relations
-            ]
+            resolved = []
```
chrisseto: Did anything change here?

share/models/meta.py
aaxelb: I removed it since it was always inferable from `url`, but I can see how it'd be useful to store it separately. If we call it `domain` it shouldn't include the scheme, though...

chrisseto: I don't think we really care about scheme. `https://orcid.org` and `http://orcid.org` are the same location. As well as `https://orcid.org/` and `https://orcid.org`.
Domains have a single way to be expressed.

share/models/validators.py
aaxelb: I thought choices [had to be](https://docs.djangoproject.com/en/1.9/ref/models/fields/#choices) a list of 2-tuples, which breaks validation because `'key' != ('key', 'label')`. If there were choices fields that were lists of strings (or grouped choices, come to think of it), it will break.

chrisseto: Whoops! I know how to django... Wouldn't you want to be validating against label in that case?

share/models/meta.py
aaxelb: The second element is there because I thought choices needed a label for each item to display in a hypothetical django form. I was thinking of adding a label to each type to put here instead of the uri, which would also make `/api/relationtypes` a bit more useful.

share/models/change.py
aaxelb: Excellent, thanks!

share/change.py
aaxelb: If I'm reading it right, the topological sort ensures that any through-node is visited after the nodes it connects, but doesn't guarantee anything about the relative order of identifiers, works, people, etc. Identifiers are topologically equivalent (as it were) to the objects they identify, but need to be visited first. Sorting the list seemed simplest, but it'll be a special case no matter what.

share/change.py
aaxelb: Oh, that's an artifact from when I made identifiers many-to-one; sometimes `next` would be given an empty sequence and raise `StopIteration`. The change should be equivalent except for handling that case gracefully.

share/change.py
chrisseto: Relevant comment from above.

> This is a subtle but very important nuance. The order that these are returned in is critically import to the disambiguation process.

>There are a few other tests that check this but I'd sleep better know that there are many assertions for it.

> See test_topological_sort_many_to_many for another example.


Identifiers will always be first as they are referenced by creativeworks or by people and they,  themselves, don't reference anything. The purpose of the resolve relations function is to ensure that creativeworks have outgoing references to anything they are connected to by a through table.


```
            --------------------------------
           /                                \
CreativeWork <-- Through --> Identifier <----


```

aaxelb: Ohh, I missed that the result of resolving those relations is used by the sort (though now it seems obvious). Yeah, the existing logic should be fine.

tests/share/changes/test_change.py
aaxelb: Fair enough, makes sense.

share/disambiguation.py
aaxelb: Ideally, yes, but (at the moment) it's possible to end up with multiple. Like if two works exist, with identifiers `I1` and `I2`, and then we harvest a work with identifiers `I1` and `I2`. Right now the new work might disambiguate on `I1`, but then `I2` will point to two works. I'm not sure if there's a better way... maybe it'd be better to merge the two existing works and then update with the new one, but we need to fix merging first, at least.

share/models/validators.py
aaxelb: I figured we'd want the normalizers to use the key (like `cites`, `is_part_of`, etc.) instead of the display label? Seems less potentially ambiguous.

share/models/meta.py
rickjohnson: I agree that 'from' and 'to' are good enough

share/models/meta.py
rickjohnson: I think it is important to store the uri to allow for additional ontologies to be used in the future.  I do have one question here in thinking how you have a particular string key per item...

For example, say 3-6 months from now we want to support an additional ontology of relationships (especially if it is being loaded from another system), and there are overlaps in relationships between the two different ones (e.g., they both have a 'cites' relationship).  I assume with a typical dictionary/hash you would only have one value per key.  So, I am wondering if the key should be changed to include something like '_spar'.   Another option might be a key pointing to a list of possible uri's (of course more complicated and maybe not something to do now).

Are you storing the key as value in the relationship table or is a link to a foreign that points to the item in the relationship table?  If it is a foreign key, than that may be good enough as you could then reference the key or value at any point.  

Thoughts?

aaxelb: When ontologies overlap, I think it makes more sense to have a key pointing to a list of possible uris, rather than having multiple keys that are semantically equivalent.

Right now the key is stored as a value in the relation table, because it seems sufficient for the queries we'll make and avoids the overhead of joining yet another table when fetching works. The list of allowed relation types is stored in a [https://github.com/aaxelb/SHARE/blob/aee324d9b5088f039634d90eba28484b89558413/share/models/relation-types.json](json file) which is (1) used for validation and (2) exposed at `/api/relationtypes` for eventual use in the curation interface. I can change the file to have a list of uris for each key, and if/when we want to support additional ontologies we could use it in the normalizer to map a uri to a key.

aaxelb:
@chrisseto I've made all the changes you requested 🐣

share/tasks.py
```diff
@@ -132,13 +132,15 @@ def run(self, normalized_id, started_by_id=None):
...
+        cs = None
         try:
             cs = ChangeSet.objects.from_graph(ChangeGraph.from_jsonld(normalized.normalized_data, extra_namespace=normalized.source.username), normalized.id)
             if cs and (normalized.source.is_robot or normalized.source.is_trusted):
                 # TODO: verify change set is not overwriting user created object
                 cs.accept()
         except Exception as e:
-            logger.exception('Failed make json patches (%d)', normalized_id)
+            if cs and cs.id:
+                cs.delete()
```
chrisseto: Can you make sure that this will delete the individual changes as well? It should just delete cascade.
Do you happen to remember why we decided against just using a transaction here?

share/models/validators.py
```diff
@@ -148,12 +148,18 @@ def validator_for(self, model):
...
 
         return JSONLDValidator.__schema_cache.setdefault(model, Draft4Validator(schema, format_checker=draft4_format_checker))
+
+    def allowed_fields_for_model(self, model):
+        excluded = {'id', 'type', 'sources', 'changes', 'same_as', 'extra'}
+        fields = model._meta.get_fields()
+        allowed_fields = [f for f in fields if f.editable and f.name not in excluded]
+        # Include one-to-many relations to models with no other relations
+        # TODO better way to exclude versions?
+        allowed_fields.extend(f for f in fields if f.one_to_many and f.name not in excluded and 'version' not in f.name and not [rf for rf in f.related_model._meta.get_fields() if rf.editable and rf.is_relation and rf.rel != f and rf.name not in excluded])
```
chrisseto: You can check the related model for the lack of a `VersionModel` attribute

share/models/identifiers.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+class CreativeWorkIdentifier(ShareObject):
+    uri = ShareURLField(unique=True)
+    host = models.TextField(editable=False)
+    scheme = models.TextField(editable=False)
+    creative_work = ShareForeignKey('AbstractCreativeWork', related_name='%(class)ss')
+
+    def save(self, *args, **kwargs):
+        f = furl(self.uri)
```
chrisseto: `furl` might not work for all the cases here. Emails, issns, and maybe others?

share/models/identifiers.py
```diff
@@ -0,0 +1,57 @@
...
+#        self.host = f.host
+#        self.scheme = f.scheme
+#        super(Identifier, self).save(*args, **kwargs)
+#
+#    class Meta:
+#        abstract = True
+
+
+class CreativeWorkIdentifier(ShareObject):
+    uri = ShareURLField(unique=True)
```
chrisseto: We'll have to document fairly thoroughly what everything here is.
Should we switch to calling them irirs, authority, and scheme? I think that's how the RFC for IRIs refer to them. I personally thing uri is a touch more clear at first glance but I technically speaking uri's don't support unicode (I think that's the main difference)

share/models/base.py
```diff
@@ -43,8 +43,11 @@ class ShareObjectMeta(ModelBase):
     }
 
     def __new__(cls, name, bases, attrs):
-        if (models.Model in bases and attrs['Meta'].abstract) or len(bases) > 1:
-            return super(ShareObjectMeta, cls).__new__(cls, name, bases, attrs)
+        try:
+            if attrs['Meta'].abstract or len(bases) > 1:
+                return super(ShareObjectMeta, cls).__new__(cls, name, bases, attrs)
+        except (KeyError, AttributeError):
```
chrisseto: What's the case here?

share/tasks.py
aaxelb: We decided against a transaction because I saw a deadlock that seemed like it might have been caused by wrapping all the disambiguation in a transaction, but now that I think about it that doesn't seem right... I'll put it in a transaction to be sure all the changes are rolled back, but we should keep an eye on it under load.

share/models/validators.py
aaxelb: Makes sense 👍

share/models/identifiers.py
aaxelb:
```
>>> f = furl('mailto:foo@bar.com')
>>> f.scheme, f.host
('mailto', 'bar.com')
>>> f = furl('urn://ISSN/1234-1234')
>>> f.scheme, f.host
('urn', 'issn')
```
I fully believe there are other cases it won't handle, but I haven't found them yet. I think it'll be fine, especially with that very nice set of shiny new IRI-parsing links.

share/models/base.py
aaxelb: Goooood question... I was trying to give all identifiers a common base class, and I guess I wanted that base class to inherit ShareObject, but only its subclasses get the magic from ShareObjectMeta. This was probably the wrong way to go about it and isn't relevant since I gave up. I'll revert it.

PR #430, opened 2016-09-22
Registration v2 [SHARE-360]

leb2dg:
Jira: https://openscience.atlassian.net/browse/SHARE-360

### TODO
* add in adding direct sources via admin interface

### Changes
* add registration endpoint
* add registration model
* add registrations to the admin interface


api/views/registration.py
```diff
@@ -0,0 +1,52 @@
...
+                            "source_preferred_metadata_prefix": "(Optional)",
+                            "source_oai": false,
+                            "source_base_url": "(Optional)",
+                            "source_disallowed_sets": "(Optional)",
+                            "source_additional_info": "(Optional)"
+                        }
+
+        Success:       200 OK
+    """
+    permission_classes = [ReadOnlyOrTokenHasScopeOrIsAuthenticated, ]
```
chrisseto: I know this isn't quite consistent but could you use a tuple here rather than a list? Tuples are immutable and imply that the value should not be changed.

api/views/registration.py
```diff
@@ -0,0 +1,52 @@
...
+    serializer_class = RegistrationFormSerializer
+
+    def get_queryset(self):
+        return RegistrationForm.objects.all()
+
+    def create(self, request, *args, **kwargs):
+        serializer = RegistrationFormSerializer(data=request.data, context={'request': request})
+        if serializer.is_valid():
+            nm_instance = serializer.save()
+            return Response({'registration_id': nm_instance.id}, status=status.HTTP_202_ACCEPTED)
```
chrisseto: Successful create requests should return 201 (Created).
202s signify that your initial data was accepted but there is further processing to be done.

api/serializers.py
```diff
@@ -11,6 +11,16 @@ class Meta:
         fields = ('id', 'source', 'app_label', 'provider_doc_id', 'data', 'sha256', 'date_seen', 'date_harvested')
 
 
+class RegistrationFormSerializer(serializers.ModelSerializer):
+    submitted_at = serializers.DateTimeField(read_only=True)
```
chrisseto: Could you order these alphabetically?

share/admin.py
```diff
@@ -179,6 +180,17 @@ class AccessTokenAdmin(admin.ModelAdmin):
     list_display = ('token', 'user', 'scope')
 
 
+class RegistrationFormAdmin(admin.ModelAdmin):
+    list_display = ('status_', 'source_name', 'submitted_at', 'direct_source', 'submitted_by')
+
+    readonly_fields = ('submitted_at', 'submitted_by',)
+
+    list_filter = ['status', 'direct_source']
```
chrisseto: Tuple

share/admin.py
```diff
@@ -179,6 +180,17 @@ class AccessTokenAdmin(admin.ModelAdmin):
     list_display = ('token', 'user', 'scope')
 
 
+class RegistrationFormAdmin(admin.ModelAdmin):
+    list_display = ('status_', 'source_name', 'submitted_at', 'direct_source', 'submitted_by')
```
chrisseto: Alphabetical ordering, please

share/migrations/0044_auto_20160923_1704.py
```diff
@@ -0,0 +1,51 @@
...
+                ('source_documentation', models.TextField(blank=True, null=True)),
+                ('source_preferred_metadata_prefix', models.TextField(blank=True, null=True)),
+                ('source_oai', models.BooleanField(default=False)),
+                ('source_base_url', models.URLField(blank=True, null=True)),
+                ('source_disallowed_sets', models.TextField(blank=True, null=True)),
+                ('source_additional_info', models.TextField(blank=True, null=True)),
+                ('submitted_by', models.ForeignKey(on_delete=django.db.models.deletion.CASCADE, to=settings.AUTH_USER_MODEL)),
+            ],
+        ),
+        migrations.AlterField(
```
chrisseto: How did this sneak in?

share/models/registration.py
```diff
@@ -0,0 +1,41 @@
+from model_utils import Choices
+
+from django.conf import settings
+from django.db import models
+from django.utils.translation import ugettext as _
+
+
+class RegistrationForm(models.Model):
```
chrisseto: This is kind of a dumb thing, `RegistationForm` is a bit misleading as `Form`s are a thing in Django apps. Maybe just call it `ProviderRegistration` or `RegistrationRequest`? I'm open to anything that doesn't have `Form` in the name 🎱

api/serializers.py
```diff
@@ -11,6 +11,16 @@ class Meta:
         fields = ('id', 'source', 'app_label', 'provider_doc_id', 'data', 'sha256', 'date_seen', 'date_harvested')
 
 
+class ProviderRegistrationSerializer(serializers.ModelSerializer):
+    status = serializers.IntegerField(read_only=True)
```
chrisseto: There's a related comment in the frontend PR. Could you return this as the string variant?

share/models/registration.py
```diff
@@ -0,0 +1,41 @@
...
+
+    STATUS = Choices(
+        (0, 'pending', _('pending')),
+        (1, 'accepted', _('accepted')),
+        (2, 'implemented', _('implemented')),
+        (3, 'rejected', _('rejected'))
+    )
+
+    status = models.IntegerField(choices=STATUS, default=STATUS.pending)
+    submitted_at = models.DateTimeField(auto_now_add=True)
```
chrisseto: Could you make the default ordering descending on `submitted_at` and add a complementary index? The admin interface should then sort them on `submitted_at` by default.

share/models/registration.py
```diff
@@ -0,0 +1,41 @@
...
+    meta_rights = models.BooleanField(default=False)
+    meta_privacy = models.BooleanField(default=False)
+    meta_sharing = models.BooleanField(default=False)
+    meta_license = models.BooleanField(default=False)
+
+    direct_source = models.BooleanField(default=False)
+
+    source_name = models.TextField()
+    source_description = models.TextField()
+    source_rate_limit = models.TextField(blank=True, null=True)
```
chrisseto: When dealing with text fields [avoid using `null=True`](https://docs.djangoproject.com/en/1.10/ref/models/fields/#null) if possible.

leb2dg: What should be used instead? It's an optional field.

share/models/registration.py
```diff
@@ -0,0 +1,41 @@
...
+
+    meta_tos = models.BooleanField(default=False)
+    meta_rights = models.BooleanField(default=False)
+    meta_privacy = models.BooleanField(default=False)
+    meta_sharing = models.BooleanField(default=False)
+    meta_license = models.BooleanField(default=False)
+
+    direct_source = models.BooleanField(default=False)
+
+    source_name = models.TextField()
```
chrisseto: We should probably put maximum lengths on all the textfields here, just in case. They can be fairly generous but draw the line somewhere. 🚒

share/models/registration.py
chrisseto: Just `blank=True`

PR #438, opened 2016-10-06
[SHARE-401] [Feature] PMC-FM metadata format

zamattiac:
# At long last, it is here
## Possible Issues
- Inconsistent structure. Every record is unique as a snowflake ❄️
- This normalizer was run successfully on 700 records, so most common inconsistencies are managed
### Ticket

https://openscience.atlassian.net/browse/SHARE-401


providers/gov/pubmedcentral/pmc/apps.py
```diff
@@ -0,0 +1,21 @@
...
+    time_granularity = False
+    harvester = PMCHarvester
+    namespaces = {
+        'http://www.openarchives.org/OAI/2.0/': None,
+        'http://jats.nlm.nih.gov/ns/archiving/1.0/': None,
+        'http://www.w3.org/2001/XMLSchema-instance': 'xsi',
+        'http://www.niso.org/schemas/ali/1.0': 'ali',
+        'http://www.w3.org/1999/xlink': 'xlink',
+        'http://www.w3.org/1998/Math/MathML': 'mml'
+    }
```
chrisseto: As this is just an extension to pubmedcentral could you add a user property to refer to the original pubmed user?
https://github.com/CenterForOpenScience/SHARE/blob/develop/providers/io/osf/preprints/apps.py

providers/gov/pubmedcentral/pmc/migrations/0001_initial.py
```diff
@@ -0,0 +1,26 @@
...
+    ]
+
+    operations = [
+        migrations.RunPython(
+            code=share.robot.RobotUserMigration('gov.pubmedcentral.pmc'),
+        ),
+        migrations.RunPython(
+            code=share.robot.RobotOauthTokenMigration('gov.pubmedcentral.pmc'),
+        ),
+        migrations.RunPython(
```
chrisseto: This should only have the `RobotScheduleMigration` the pubmed user already exists.

providers/gov/pubmedcentral/pmc/normalizer.py
```diff
@@ -0,0 +1,243 @@
...
+
+    def get_contributors(self, ctx, type):
+        results = []
+
+        if type == 'person':
+            for contributor in ctx:
+                try:
+                    contributor['name']
+                    results.append(contributor)
+                except:
```
chrisseto: Don't use bare except block, they will catch all exceptions, even keyboard interrupts.
Instead just catch `Exception`

However if you're just trying to catch a `KeyError`  just catch that or do a check ahead of time

```python
if 'name' in contributor:
...
```

PR #440, opened 2016-10-06
[SHARE-380] Allow objects to transition generic types

aaxelb:
Subset of changes in #429, so we don't have to wait for 2.0.0

### DIFF
--- a/share/change.py
+++ b/share/change.py
@@ -37,6 +37,16 @@ class ChangeNode:
     def instance(self):
         return self.__instance
 
+    @instance.setter
+    def instance(self, instance):
+        if instance:
+            self.id = instance.pk
+            self.type = instance._meta.model_name.lower()
+            if self.type != self.new_type and self.new_type == 'creativework':
+                self.new_type = self.type
+            self.__refs.append((self.id, self.type))
+        self.__instance = instance
+
     @property
     def is_blank(self):
         # JSON-LD Blank Node ids start with "_:"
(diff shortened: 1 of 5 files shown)


chrisseto:
Is there a reason to use `new_type`? it looks like you could get away just find if you keep `self.__refs` up to date

aaxelb:
That makes sense... It just means doing the `creativework` check at the last minute. Something like
```python
new_type = self.__refs[-1][1]
if new_type != self.type and new_type != 'creativework':
    ret['@type'] = new_type
```

chrisseto:
I was thinking:
```python
@instance.setter
def instance(self, instance):
    if instance:
        self.id = instance.pk
        if instance._meta.model_name.lower() != 'creativework':
             self.type = instance._meta.model_name.lower()
             self.__refs.append((self.id, self.type))
    self.__instance = instance
```

Sounds like there is something I'm not thinking of

aaxelb:
I was thinking the special case is when the submitted `@type` is 'creativework'. When the existing instance is `creativework`, it's just like any other type. That is, you can transition `creativework => article` and `article => preprint`, but if you try `preprint => creativework` it'll just stay `preprint`. Is that what we want?

An annoying thing is that this change's `type` should be the type of the existing object, so `Change.target` finds the right thing, while any nodes that reference this one should use the new type, since they're resolved after the transition. Making `Change._resolve_ref` use the concrete model for non-blank ids would resolve that problem, though...

chrisseto:
> Making Change._resolve_ref use the concrete model for non-blank ids would resolve that problem, though...

That sounds like the best solution IMO. There will probably only be good things that come from that change, especially if we plan to do more with transitioning types. TypedModels should handle loading the proxies correctly even if they are loaded from the concrete class.

PR #441, opened 2016-10-06
[SHARE-388] [Feature] Update date parser

leb2dg:
* Improve date parser link
    * add upper and lower bounds
    * `None` causes `ValueError`
    * uses dateutil instead of Arrow to parse dates
    * breaks on bad dates
* Arrow --> Pendulum
* Add tests for date parser

### DIFF
--- a/share/normalize/links.py
+++ b/share/normalize/links.py
@@ -287,8 +287,8 @@ class DateParserLink(AbstractLink):
                 raise ValueError('{} is before the lower bound {}.'.format(obj, self.LOWER_BOUND.isoformat()))
             if date > self.UPPER_BOUND:
                 raise ValueError('{} is more than 100 years in the future.'.format(obj))
-            return date.isoformat()
-        return None
+            return date.in_tz('UTC').isoformat()
+        raise ValueError('{} is not a valid date.'.format(obj))
 
 
 class LanguageParserLink(AbstractLink):
--- a/share/normalize/links.py
+++ b/share/normalize/links.py
@@ -5,10 +5,11 @@ import json
 import logging
 import re
 import threading
+import dateutil
 
 import xmltodict
 
-import arrow
+import pendulum
 
 from lxml import etree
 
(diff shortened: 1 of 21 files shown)


share/normalize/links.py
```diff
@@ -275,9 +276,18 @@ def execute(self, obj):
...
 
 class DateParserLink(AbstractLink):
+    LOWER_BOUND = pendulum.create(1200, 1, 1)
+    UPPER_BOUND = pendulum.now().add(years=100)
+    DEFAULT = pendulum.create(2016, 1, 1)
+
     def execute(self, obj):
         if obj:
-            return arrow.get(obj).to('UTC').isoformat()
+            date = dateutil.parser.parse(obj, default=self.DEFAULT)
```
chrisseto: Force the date into UTC here. If there isn't a timezone assume it is already UTC

tests/share/normalize/test_links.py
```diff
@@ -1,8 +1,35 @@
...
 
-from share.normalize.links import DOILink
-from share.normalize.links import OrcidLink
+import calendar
+
+from share.normalize.links import DOILink, OrcidLink, DateParserLink
+
+
+@pytest.mark.parametrize('date, result', [
+    (None, None),
```
chrisseto: `None` should raise an exception of some kind. We'll want to know if a normalizer is unknowingly passing along `None` values.

tests/share/normalize/test_links.py
```diff
@@ -1,8 +1,35 @@
...
+    ('0059-11-01T00:00:00Z', ValueError('0059-11-01T00:00:00Z is before the lower bound 1200-01-01T00:00:00+00:00.')),
```
chrisseto: Could you add a test case or two for coercing other time zones into UTC?
Maybe one for an invalid time zone as well?

share/normalize/links.py
chrisseto: Tests are fine!

leb2dg:
@chrisseto, comments (finally) addressed

PR #442, opened 2016-10-10
[SHARE-418] [Feature] Remove source versions tables

chrisseto:


chrisseto:
This is being handled in release/2.0.0

PR #448, opened 2016-10-12
[SHARE-423][2.0.0] Load relation types

aaxelb:
* Initialize work and entity relation types on `./up.sh`, using `addrelationtypes` command.
* Squash all the migrations so `up` works nicely.
* Rename `NodeSorter` to `TopographicalSorter`, make it generic, and move it to `share.util` so it can sort relation types, too


aaxelb:
Replaced with #454

PR #449, opened 2016-10-12
Add collaboration groups

chrisseto:


tests/share/models/factories.py
```diff
@@ -0,0 +1,124 @@
...
+        self.change.save()
+
+
+class TypedShareObjectFactory(ShareObjectFactory):
+    class Meta:
+        abstract = True
+
+    @factory.lazy_attribute
+    def type(stub):
+        return random.choice([m.label.lower() for m in stub._LazyStub__model_class._meta.model._meta.concrete_model._meta.proxied_children])
```
aaxelb: 3 _metas deep! Crazy.

tests/share/models/factories.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+class EntityFactory(TypedShareObjectFactory):
+    name = factory.Faker('company')
+    given_name = factory.Faker('first_name')
+    family_name = factory.Faker('last_name')
```
aaxelb: Does this not cause an error for non-person entities? Seems like it should.

chrisseto:
I'm not a huge fan of the naming on any of these as they aren't super clear. Whether or not you appear on the "top level" is entirely dependent on the `bibliographic` flag. `Collaboration` is the default bucket for the moment.

aaxelb:
Oh right, `bibliographic`, that makes sense. Maybe the base should be `AbstractContribution`, for clarity and to parallel the other typed models? Otherwise, I think the names are okay. `Host` is kinda weird, but I'm not sure what would be clearer.

PR #480, opened 2016-10-19
[Feature][2.0.0] OAI normalizer

aaxelb:
- OAI Normalizer! Updated to the new data model.
 - OAI sources can have a `type_map` in their config, a dictionary that maps from a `dc:type` value to the name of a creative work type. It will be used to set the type of normalized works, with `emitted_type` as the default/fallback.
- Fixed a smattering of bugs unearthed by testing against various sources.
- Add `makechanges` management command, to manually run `MakeJsonPatches` for a normalized datum.
- Add some useful normalizer things:
 - `GuessAgentType`, to guess the type of agent (person, institution, organization) based on their name
 - `OAILink`, used by `IRI` for OAI identifiers like `oai:nature.com:foo`
 - `ArXivLink`, used by `IRI` for ArXiv identifiers like `arXiv:1010.22981`


share/normalize/links.py
```diff
@@ -533,7 +543,7 @@ def execute(self, obj):
         mapped = self.MAPPING.get(obj.lower())
 
         if not mapped:
-            logger.warning('No synonyms found for term "%s"', obj)
+            logger.debug('No synonyms found for term "%s"', obj)
```
chrisseto: Thank you

share/normalize/links.py
```diff
@@ -832,5 +892,52 @@ def execute(self, obj):
                 break
 
         if not final[0]:
-            raise ValueError('\'{}\' could not be identified as an Identifier.'.format(obj))
```
chrisseto: When would you want to not fail here? As Identifiers are fairly important, I'd rather not make them implicitly fire and forget.

share/normalize/links.py
```diff
@@ -832,5 +892,52 @@ def execute(self, obj):
...
+            return {'IRI': None}
         return final[0]().execute(obj)
+
+
+class GuessAgentTypeLink(AbstractLink):
+    """
+    When executed on the name of an agent, guess the agent's type.
+    """
+
+    ORGANIZATION_KEYWORDS = (
```
chrisseto: Would it be more efficient to just compile all of these into a single regex?

aaxelb: Yeah, probably, it'll be used a lot. Good call.

PR #497, opened 2016-11-01
[2.0.0] Update datacite normalizer

leb2dg:
### Changes
* Update to work with release 2.0.0
* Fix harvester pendulum problem
* Handle both v3 and v4 of Datacite's schema
* Add a few more work relations


providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
+    """
+    Returns agent type based on contributor type.
+    """
+    is_not_person = try_contributor_type(agent, NOT_PEOPLE_TYPES)
+    is_person = try_contributor_type(agent, PEOPLE_TYPES)
+
+    if is_person:
+        return 'Person'
+    elif is_not_person:
+        return 'Organization'
```
aaxelb: Could any of the `NOT_PEOPLE_TYPES` be an institution? Not sure if `HostingInstitution` means it the same way we do.

providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
+    schema = 'Creator'
 
-    name = ctx
+    order_cited = ctx('index')
 
 
-class Identifier(Parser):
-    base_url = tools.Try(ctx['@schemeURI'])
-    url = tools.Join(tools.Concat(tools.Try(ctx['@schemeURI']), tools.Try(ctx['#text'])), joiner='/')
+class HostRelation(Parser):
```
aaxelb: Is `HostRelation` ever used?

providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
+    name = tools.Try(ctx.awardTitle)
+    description = tools.Try(ctx.awardNumber)
+    uri = tools.Try(ctx.awardURI)
+
+
+class FunderRelation(Parser):
+    schema = 'Funder'
 
+    agent = tools.Delegate(FunderAgent, ctx)
+    award = tools.Delegate(Award, ctx)
```
aaxelb: Should be `awards`, a many-to-many to `Award` through `ThroughAwards`

providers/org/datacite/normalizer.py
```diff
@@ -7,8 +6,55 @@
...
+    'Editor',
+    'ProjectLeader',
+    'ProjectManager',
+    'ProjectMember',
+    'RelatedPerson',
+    'Researcher',
+    'Supervisor',
+    'WorkPackageLeader'
+)
+CHECK_TYPES = (
```
aaxelb: Is `CHECK_TYPES` ever used?

providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
+        if inverse and relation in INVERSE_RELATIONS:
+            results.append(option)
+        elif not inverse and relation in RELATIONS:
+            results.append(option)
+    return results
+
+
+def get_relation_type(relation_type):
+    normalized_relation = RELATION_MAP[relation_type]
+    return normalized_relation if normalized_relation else 'WorkRelation'
```
aaxelb: Could be simplified:
```
return normalized_relation or 'WorkRelation'
```
I like short circuiting.

providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
+        tools.Delegate(AgentIdentifier),
+        tools.Try(
+            tools.RunPython(
+                force_text,
+                ctx.nameIdentifier
+            )
+        )
     )
-    type = tools.RunPython('lower', tools.Try(ctx['@relatedIdentifierType']))
+    related_agents = tools.Map(tools.Delegate(AffiliatedAgent), tools.Concat(tools.Try(
```
aaxelb: Needs to go through an agent relation. `IsAffiliatedWith`, maybe?

providers/org/datacite/normalizer.py
```diff
@@ -28,243 +74,402 @@ def force_text(data):
...
-class ContributorOrganization(Parser):
-    schema = 'Organization'
+class PublisherAgent(Parser):
+    schema = tools.GuessAgentType(ctx, default='organization')
 
-    name = ctx.contributorName
+    name = ctx
 
     class Extra:
         contributor_type = tools.Try(ctx.contributorType)
```
aaxelb: `ctx` is the publisher name here, so it'll always be a string, and this `Try` emits a warning for every record. I think you can get rid of it.

providers/org/datacite/normalizer.py
leb2dg: This needs to be plural; `award` --> `awards`

PR #502, opened 2016-11-02
[2.0.0] Updated DataOne normalizer

aaxelb:
- Update DataOne normalizer for 2.0.0
- Add `ARKLink`, since many DataOne identifiers are [ARKs](https://en.wikipedia.org/wiki/Archival_Resource_Key)
- Handle URLs that start with `www.` instead of a scheme, because that happens
- Add `Documents` work relation, but leave the migration off because it's covered by #497

### DIFF
--- a/share/normalize/links.py
+++ b/share/normalize/links.py
@@ -409,7 +409,7 @@ class TryLink(AbstractLink):
         except self._exceptions:
             return self._default
         except TypeError as err:
-            logger.warning('TypeError: {}. When trying to access {}'.format(err, self._chain))
+            logger.warning('TypeError: {}. When trying to access {}'.format(err, self._chain.chain()))
             return self._default
         return self.__anchor.run(val)
 
@@ -469,6 +469,9 @@ class XPathLink(AbstractLink):
         if len(elems) == 1 and not isinstance(self._next, (IndexLink, IteratorLink)):
             return elems[0]
         return elems
+
+    def __repr__(self):
+        return '<{}({!r})>'.format(self.__class__.__name__, self._xpath)
 
 
 class DelegateLink(AbstractLink):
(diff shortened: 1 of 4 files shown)


share/normalize/links.py
```diff
@@ -764,10 +768,14 @@ class URLLink(AbstractIRILink):
     def hint(cls, obj):
         if cls.URL_RE.search(obj) is not None:
             return 0.25
+        if obj[:4].lower() == 'www.':
```
chrisseto: While you're at it would you mind adding `ww2.` as well? I've seen that floating around some places.

share/normalize/links.py
```diff
@@ -840,6 +848,30 @@ def _parse(self, obj):
         }
 
 
+class ARKLink(AbstractIRILink):
+    # https://en.wikipedia.org/wiki/Archival_Resource_Key
+    # https://wiki.ucop.edu/download/attachments/16744455/arkspec.pdf
+
+    ARK_SCHEME = 'ark'
+    ARK_RE = re.compile(r'\bark:/(\d+)(/\S+)', flags=re.I)
```
chrisseto: 🐶

share/normalize/links.py
```diff
@@ -840,6 +848,30 @@ def _parse(self, obj):
         }
 
 
+class ARKLink(AbstractIRILink):
```
chrisseto: Would we want to express arks as URL like we do with DOIs?
I guess the better question is, do we get enough information to?

share/normalize/links.py
aaxelb: `ww2.` or `www2.`? I've seen the latter, but I guess neither would surprise me.

chrisseto: Latter, sorry. 🏚

share/normalize/links.py
aaxelb: We don't get enough information to. They're usually just `ark:/1234/blah/blah`, and there's no single host that can resolve any ARK, unlike DOIs. They're useful as identifiers, but not locators.

chrisseto: 👍

share/normalize/links.py
aaxelb: Done 🌬

PR #504, opened 2016-11-03
[2.0.0] Update USGS normalizer

aaxelb:
Also:

- Add `SourceID` to normalizer tools.
  - For sources that give objects IDs that aren't URI-like or globally unique, make a URN that `IRI` will recognize.
  - Change `OAILink` to `URNLink`, which can accept URNs with scheme `urn` or `oai`
  - e.g. `SourceID(ctx.contributorID)` in the `gov.usgs` normalizer yields something like `urn:share:gov.usgs:1234`
- Make sure `Context().parser` is always reset when leaving a parser, even if leaving via exception
  - Allows wrapping `Try` around `Delegate`.

### DIFF
--- a/share/normalize/links.py
+++ b/share/normalize/links.py
@@ -24,7 +24,7 @@ from share.util import DictHashingDict
 logger = logging.getLogger(__name__)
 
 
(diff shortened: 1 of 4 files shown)


share/normalize/links.py
```diff
@@ -922,3 +930,10 @@ def __init__(self, func):
 
     def execute(self, obj):
         return list(filter(self._func, obj))
+
+
+class SourceIDLink(AbstractLink):
+    URN_FORMAT = 'urn:share:{source}:{id}'
+
+    def execute(self, obj):
+        return self.URN_FORMAT.format(source=Context()._config.label, id=obj)
```
chrisseto: Should id be url encoded? For some of the ids I've just been sending back urls 😨

aaxelb: Oh yeah, meant to do that. Done 👍

PR #522, opened 2016-11-17
[SHARE-438][Feature] Generate endpoints for generated models

leb2dg:
### Purpose
Generate endpoints for generated models so we don't have to update the api every time a model changed.

screen shot 2016-12-09 at 1 18 58 pm

screen shot 2016-12-09 at 10 15 21 am

### Ticket
https://openscience.atlassian.net/browse/SHARE-438


chrisseto:
Looks great! Could you remove `share/serializers.py` and add a couple test cases?

leb2dg:
I import `BaseSerializer` from `share/serializers.py`. Should I move that or just delete the extra serializers and not the whole file?

chrisseto:
@leb2dg I'd go with move it

api/views/share.py
```diff
@@ -59,102 +61,34 @@ def rawdata(self, request, pk=None):
...
-    serializer_class = serializers.WorkRelationSerializer
-    queryset = serializer_class.Meta.model.objects.all().select_related('extra')
+        assert lookup_url_kwarg in self.kwargs, (
+            'Expected view %s to be called with a URL keyword argument '
+            'named "%s". Fix your URL conf, or set the `.lookup_field` '
+            'attribute on the view correctly.' %
+            (self.__class__.__name__, lookup_url_kwarg)
+        )
 
+        decoded_pk = IDObfuscator.decode(self.kwargs[lookup_url_kwarg])[1]
```
chrisseto: Would it be possible to check that the model requested here is appropriate for this endpoint and then send back a 400 if it is not?

`decode` can also raise exceptions.  Could you make a custom exception for `InvalidID`, or something along those lines, and handle it here. Either a 400 or 404 would be fine, just depends on how nice you're feeling.

api/urls.py
```diff
@@ -15,48 +22,97 @@ def register_route(prefix, viewset):
...
+        self.fields.update({
+            'type': fields.TypeField(),
+        })
+
+    class Meta:
+        links = ('versions', 'changes', 'rawdata')
+
+    # http://stackoverflow.com/questions/27015931/remove-null-fields-from-django-rest-framework-response
+    def to_representation(self, instance):
+        def not_none(value):
```
chrisseto: This function serves no purpose. `lambda x: x[1] is not None` is much easier to follow.

api/urls.py
```diff
@@ -15,48 +22,97 @@ def register_route(prefix, viewset):
...
-# # identifier routes
-# register_route(r'agentidentifier', views.AgentIdentifierViewSet)
-# register_route(r'workidentifier', views.WorkIdentifierViewSet)
-
-# # relation routes
-# register_route(r'agentrelation', views.AgentRelationViewSet)
-# register_route(r'workrelation', views.WorkRelationViewSet)
-# register_route(r'contribution', views.ContributionViewSet)
-# register_route(r'award', views.AwardViewSet)
+class BaseShareSerializer(serializers.ModelSerializer):
```
chrisseto: This does not belong in `urls.py`

api/urls.py
```diff
@@ -15,48 +22,97 @@ def register_route(prefix, viewset):
...
+
+class EndpointGenerator:
+
+    def __init__(self):
+        subclasses = models.ShareObject.__subclasses__()
+
+        generated_endpoints = []
+        for subclass in subclasses:
+            if not (subclass._meta.proxied_children and subclass == subclass._meta.concrete_model):
+                generated_endpoints += [subclass]
```
chrisseto: `.append`

api/urls.py
```diff
@@ -15,48 +22,97 @@ def register_route(prefix, viewset):
...
+
+
+class EndpointGenerator:
+
+    def __init__(self):
+        subclasses = models.ShareObject.__subclasses__()
+
+        generated_endpoints = []
+        for subclass in subclasses:
+            if not (subclass._meta.proxied_children and subclass == subclass._meta.concrete_model):
```
chrisseto: When testing for the equality of two classes `is` is preferred as classes (and primitives) are always the same object.

```python
In [1]: dict == dict
Out[1]: True

In [2]: dict is dict
Out[2]: True

In [3]: dict is type({})
Out[3]: True

In [4]: type({}) is type({})
Out[4]: True

In [5]: {} == {}
Out[5]: True

In [6]: {} is {}
Out[6]: False
```

leb2dg:
@chrisseto, where should `BaseSerializer` live?

PR #533, opened 2016-12-06
[Feature] OAIProvider Generator

zamattiac:
- Updated readthedocs
- CSIC provided as an example (should be identical to the other ticket)

# Actions
- Creates TLD if necessary
- Prompts adding to gitignore, settings, etc.
- Fetches favicon and runs sample harvest
- Makes migration
- Template-fills provider folder, apps, __init__ using CookieCutter
- Runs flake8


chrisseto:
@zamattiac Tests are broken

PR #534, opened 2016-12-07
[Feature][SHARE-416] Retractions!

aaxelb:
- Add `Retraction` work type and `Retracts` work relation type
- Update PMC normalizer to recognize retractions and work relations
- Add retractions to elasticsearch
  - Searchable `retracted` boolean flag
  - Unindexed list of retractions in result payload
- Various incidental changes
  - Have `./bin/share normalize -d` skip hidden files
  - Add some shell_plus utils: `get_raws` and `print_raws`, which get/print all the raw data for an object or object ID
  - Allow harvesting works by provider ID (implemented for OAI sources only): `./manage.py harvest gov.pubmedcentral.pmc --ids oai:pubmedcentral.nih.gov:4512020`


share/disambiguation.py
```diff
@@ -131,9 +134,12 @@ def _instance_for_node(self, node):
                 return None
             if len(found) == 1:
                 return found[0]
-            if all_query.children or any('__' in str(query) for query in queries):
+            if all_query.children:
                 logger.warning('Multiple %ss returned for %s (The main query) bailing', concrete_model, all_query)
                 break
+            if all('__' in str(query) for query in queries):
```
chrisseto: Why is this changed to `all`?  If we have 2 identifiers and a name this case won't be hit.

aaxelb: Yeah, that was exactly the situation that made me change it. I had an organization with 2 identifiers and a name, and because there was a person with the same name, the query had two results. With `all` here, it'll narrow the search by type in the next loop iteration and correctly disambiguate to the matching organization. If the query didn't use a field on the model, though, there'd be no point narrowing by type, which is why `all` seemed right.

PR #535, opened 2016-12-08
[SHARE-354] [Feature] Cascade deletes

chrisseto:


db/deletion.py
```diff
@@ -0,0 +1,25 @@
...
+class DatabaseOnDelete:
+
+    def __init__(self, clause='NO ACTION'):
+        self.clause = clause
+
+    def __call__(self, collector, field, sub_objs, using):
+        pass
+
+    def __eq__(self, other):
+        return DO_NOTHING == other or (isinstance(other, DatabaseOnDelete) and other.clause == self.clause)
```
aaxelb: ~~Why is `DO_NOTHING` always equal?~~
Oh I see, [part of that dirty hack](https://github.com/django/django/blob/master/django/db/models/deletion.py#L148). I think that's worth a comment?

share/migrations/0007_auto_20161122_1810.py
```diff
@@ -26,7 +26,7 @@ class Migration(migrations.Migration):
         migrations.AlterField(
```
aaxelb: Are there any changes in this migration? Looks like it's just choices in a different order.

share/models/base.py
```diff
@@ -50,9 +51,10 @@ def __new__(cls, name, bases, attrs):
 
         version_attrs = {}
         for key, val in attrs.items():
-            if isinstance(val, models.Field) and val.unique:
+            if isinstance(val, models.Field) and (val.unique or val.db_index):
                 val = copy.deepcopy(val)
                 val._unique = False
+                val.db_index = False
```
aaxelb: 👍 I wonder how big our useless version indexes are...

share/models/core.py
```diff
@@ -183,9 +186,9 @@ def store_data(self, doc_id, data, source, app_label):
         )
 
         if created:
-            logger.debug('Newly created RawData for document {} from {}'.format(doc_id, source))
+            logger.debug('Newly created RawData for document %s from %s', doc_id, source)
```
aaxelb: Is there a benefit to using `%s` and logger args? Should we be doing that everywhere?

share/models/fields.py
```diff
@@ -97,66 +105,137 @@ def validate(self, value, model_instance):
...
-        super().__init__(model, **kwargs)
-
     def contribute_to_class(self, cls, name, **kwargs):
-        actual = self.__class__.mro()[2](self.remote_field.model, **self._kwargs)
+        actual = self.PRETENDING_TO_BE(self.remote_field.model, **self._get_kwargs(cls))
         actual.contribute_to_class(cls, name, **kwargs)
 
         if isinstance(self.remote_field.model, str):
-            version = self.__class__.mro()[2](self.remote_field.model + 'Version', editable=False, **{**self._kwargs, 'db_index': False})
+            version_model = self.remote_field.model + 'Version'
```
aaxelb: Would it be worth checking if the named model is a ShareObject subclass, and return if not?

share/models/fields.py
```diff
@@ -97,66 +105,137 @@ def validate(self, value, model_instance):
...
+    PRETENDING_TO_BE = models.ForeignKey
 
+    def __init__(self, *args, **kwargs):
+        # Default to delete cascade
+        kwargs.setdefault('on_delete', DATABASE_CASCADE)
+        super().__init__(*args, **kwargs)
 
-class TypedManyToManyField(ShareRelatedField, models.ManyToManyField):
+
+def create_many_to_many_intermediary_model(field, klass):
```
aaxelb: Does the generated through model differ from Django's in any way other than `on_delete=DATABASE_CASCADE`?

share/models/base.py
```diff
@@ -22,7 +23,7 @@
 
 class ShareObjectVersion(models.Model):
     action = models.TextField(max_length=10)
-    persistent_id = models.PositiveIntegerField()  # Must match the id of ShareObject
+    # persistent_id = models.PositiveIntegerField()  # Must match the id of ShareObject
```
aaxelb: Delete this?

share/models/fields.py
```diff
@@ -97,66 +105,137 @@ def validate(self, value, model_instance):
...
+            db_tablespace=field.db_tablespace,
+            db_constraint=field.remote_field.db_constraint,
+            on_delete=DATABASE_CASCADE,
+        )
+    })
+
+
+class TypedManyToManyField(models.ManyToManyField):
+
+    # PRETENDING_TO_BE = models.ManyToManyField
```
leb2dg: Delete?

tests/share/models/factories.py
```diff
@@ -144,3 +144,38 @@ class TagFactory(ShareObjectFactory):
...
+
+class ThroughTagsFactory(ShareObjectFactory):
+    tag = factory.SubFactory(TagFactory)
+    creative_work = factory.SubFactory(AbstractCreativeWorkFactory)
+
+    class Meta:
+        model = models.ThroughTags
+
+
+class WorkIdentifierFactory(ShareObjectFactory):
```
leb2dg: Are these duplicate factories purposely separate or meant to replace the ones in https://github.com/CenterForOpenScience/SHARE/blob/9aac202c7315442b97b58a275e9a7ea83f661e62/tests/share/normalize/factories.py?

share/models/fields.py
```diff
@@ -97,66 +105,137 @@ def validate(self, value, model_instance):
...
-
-
-class ShareOneToOneField(ShareRelatedField, models.OneToOneField):
-    def __init__(self, model, **kwargs):
-        self._kwargs = kwargs
-        super().__init__(model, **kwargs)
-
     def contribute_to_class(self, cls, name, **kwargs):
-        actual = self.__class__.mro()[2](self.remote_field.model, **self._kwargs)
+        actual = self.PRETENDING_TO_BE(self.remote_field.model, **self._get_kwargs(cls))
```
leb2dg: Is `PRETENDING_TO_BE` setting versions to the django model instead of the SHARE model?

tests/share/models/test_cascade.py
```diff
@@ -0,0 +1,300 @@
...
+
+        assert models.WorkIdentifier.objects.exclude(id=None).count() == 0
+        assert models.AbstractCreativeWork.objects.exclude(id=None).count() == 0
+        assert models.WorkIdentifierVersion.objects.exclude(id=None).count() == 0
+        assert models.AbstractCreativeWorkVersion.objects.exclude(id=None).count() == 0
+
+    def test_many_to_many(self):
+        work = factories.AbstractCreativeWorkFactory()
+        for _ in range(10):
+            factories.ThroughTagsFactory(creative_work=work)
```
aaxelb: This morning I got around to checking your branch out and testing locally -- looks like this factory sometimes makes duplicate tags:
```
django.db.utils.IntegrityError: duplicate key value violates unique constraint "share_tag_name_key"
DETAIL:  Key (name)=(corrupti) already exists.

File: share/migrations/0007_auto_20161122_1810.py (hunk 2 above)
Comment (chrisseto) [inline share/migrations/0007_auto_20161122_1810.py]: The choices are in a different order, because they are sorted now. If you run `makemigrations` over and over right now you'll get changes to types each time.
This mitigates that issue.

File: share/models/base.py (hunk 3 above)
Comment (chrisseto) [inline share/models/base.py]: They're actually non existent. I add this in because one of them some how slipped through.

File: share/models/core.py (hunk 4 above)
Comment (chrisseto) [inline share/models/core.py]: Ideally yes? It saves a touch of time as `.format` is always run. Using the `%s` will only be formatted if the log level matches.
The most important place to use this is when logging query sets and the like

File: share/models/fields.py (hunk 5 above)
Comment (chrisseto) [inline share/models/fields.py]: I think I just wanted to avoid the circular import. if it's a string there's not that much we can do here, I don't think.

File: share/models/fields.py (hunk 6 above)
Comment (chrisseto) [inline share/models/fields.py]: Just that. But there doesn't seem to be any other way to override it 👎

File: share/models/fields.py (hunk 10 above)
Comment (chrisseto) [inline share/models/fields.py]: Yep, `PRETENDING_TO_BE` is the django field that will be delegated to.

File: tests/share/models/factories.py (hunk 9 above)
Comment (chrisseto) [inline tests/share/models/factories.py]: On purpose, they build different things. The one in normalize builds dictionaries. This one builds models.

File: share/models/base.py [hunk 12]
```diff
@@ -83,7 +86,7 @@ def __new__(cls, name, bases, attrs):
             **{k: v() for k, v in cls.share_attrs.items()},
             'VersionModel': version,
             'same_as': fields.ShareForeignKey(name, null=True, related_name='+'),
-            'version': models.OneToOneField(version, editable=False, related_name='%(app_label)s_%(class)s_version'),
+            'version': models.OneToOneField(version, editable=False, related_name='%(app_label)s_%(class)s_version', on_delete=DATABASE_CASCADE),
             'sources': models.ManyToManyField(settings.AUTH_USER_MODEL, related_name='source_%(class)s', editable=False),
```
chrisseto: Oh poo, I though I caught that one. Thanks!

PR #542, opened 2016-12-14
[SHARE-481][Feature] Update NIH & NSFAwards normalizers

leb2dg:


chrisseto:
I believe this got merged in a very strange way. @leb2dg ?

leb2dg:
@chrisseto, I don't think it's been merged

share/normalize/utils.py
```diff
@@ -36,3 +36,7 @@ def format_address(self, address1='', address2='', city='', state_or_province=''
         return '{}, {}'.format(city, state_or_province)
 
     return address1
+
+
+def str_to_int(amount):
```
chrisseto: This might be better off as a link?

`Int(my.probably.an.number.value)`

I don't really have a strong opinion one way or the other.

providers/gov/nih/normalizer.py
```diff
@@ -214,15 +280,52 @@ def maybe_org(self, obj):
         return obj
 
     def get_organization_ctx(self, ctx):
+        # TODO lauren: filter NIL
```
chrisseto: ?

providers/gov/nih/normalizer.py
```diff
@@ -12,6 +14,21 @@ def filter_nil(obj):
         return obj
```
chrisseto: I can't actually comment on the correct line, but could you fix the indentation here?

share/normalize/utils.py
leb2dg: Ah, I did consider that. It probably would be better since all normalizers adding amount to award will have to convert the string to an int. Will update.

providers/gov/nih/normalizer.py
```diff
@@ -214,15 +280,52 @@ def maybe_org(self, obj):
...
         return org_ctx
+
+    def get_pi(self, ctx, primary=True):
+        '''
+            <PI>
+                <PI_NAME>VIDAL, MARC  (contact)</PI_NAME>
+                <PI_ID>2094159 (contact)</PI_ID>
+            </PI>
+        '''
+        pi_list = ctx['PIS']['PI'] if type(ctx['PIS']['PI']) is list else [ctx['PIS']['PI']]
```
chrisseto: `isinstance(ctx..., list)`

providers/gov/nih/normalizer.py
```diff
@@ -214,15 +280,52 @@ def maybe_org(self, obj):
...
+        org_ctx = self.get_organization_ctx(ctx)
+        # only one or none so primary is that one
+        if len(pi_list) <= 1 and primary:
+            pi_list[0]['org_ctx'] = org_ctx
+            return pi_list
+        # more than one, get the primary
+        elif len(pi_list) > 1 and primary:
```
chrisseto: No need to use `elif` if you're returning in the previous if statement

From (Effective Go)[https://golang.org/doc/effective_go.html#if]

>In the Go libraries, you'll find that when an if statement doesn't flow into the next statement—that is, the body ends in break, continue, goto, or return—the unnecessary else is omitted.

> This is an example of a common situation where code must guard against a sequence of error conditions. The code reads well if the successful flow of control runs down the page, eliminating error cases as they arise. Since error cases tend to end in return statements, the resulting code needs no else statements.

```golang
f, err := os.Open(name)
if err != nil {
    return err
}
d, err := f.Stat()
if err != nil {
    f.Close()
    return err
}
codeUsing(f, d)
```

providers/gov/nih/normalizer.py
```diff
@@ -214,15 +280,52 @@ def maybe_org(self, obj):
...
+            for pi in pi_list:
+                if '(contact)' in pi['PI_NAME']:
+                    pi['PI_NAME'] = re.sub(r'(\(contact\))', '', pi['PI_NAME']).strip()
+                    pi['PI_ID'] = re.sub(r'(\(contact\))', '', pi['PI_ID']).strip()
+                    pi['org_ctx'] = org_ctx
+                    primary_pi.append(pi)
+                    break
+            return primary_pi
+        # more than one, get the non-primary
+        elif len(pi_list) > 1 and not primary:
```
chrisseto: `if pi_list and not primary`

providers/gov/nih/normalizer.py
```diff
@@ -214,15 +280,52 @@ def maybe_org(self, obj):
...
+        pi_list = ctx['PIS']['PI'] if type(ctx['PIS']['PI']) is list else [ctx['PIS']['PI']]
+        org_ctx = self.get_organization_ctx(ctx)
+        # only one or none so primary is that one
+        if len(pi_list) <= 1 and primary:
+            pi_list[0]['org_ctx'] = org_ctx
+            return pi_list
+        # more than one, get the primary
+        elif len(pi_list) > 1 and primary:
+            primary_pi = []
+            for pi in pi_list:
```
chrisseto: If you're looking for a single result `next(...)` might be a bit nicer to read. (Just remember to catch the `StopIterationError`)

It also makes the error/null case handling a bit more explicit. This block looks like it could return a list from `0` to `n` at first glance.

providers/gov/nih/normalizer.py
leb2dg: This should return `None` if the `len(pi_list) == 1 and not primary`

providers/gov/nih/normalizer.py
```diff
@@ -7,9 +9,24 @@
...
+        return None
+    return obj
+
+
+def format_org_address(doc):
+    org_city = doc.get('ORG_CITY', '')
+    org_state = doc.get('ORG_STATE', '')
+    org_zipcode = doc.get('ORG_ZIPCODE', '')
+    org_country = doc.get('ORG_COUNTRY', '')
+    if not org_city and not org_state and not org_zipcode and not org_country:
```
chrisseto: Don't bother changing it but `not any((org_city, org_state, org_zipcode, org_country)):` is a touch cleaner

providers/gov/nih/normalizer.py
```diff
@@ -215,14 +281,52 @@ def maybe_org(self, obj):
...
+        pi_list = ctx['PIS']['PI'] if isinstance(ctx['PIS']['PI'], list) else [ctx['PIS']['PI']]
+        org_ctx = self.get_organization_ctx(ctx)
+        # only one or none so primary is that one
+        if len(pi_list) <= 1 and primary:
+            pi_list[0]['org_ctx'] = org_ctx
+            return pi_list
+        # more than one, get the primary
+        if len(pi_list) > 1 and primary:
+            pi_itr = iter(pi_list)
+            while True:
```
chrisseto: What is the purpose of this loop?

providers/gov/nih/normalizer.py
```diff
@@ -215,14 +281,52 @@ def maybe_org(self, obj):
...
+        '''
+        pi_list = ctx['PIS']['PI'] if isinstance(ctx['PIS']['PI'], list) else [ctx['PIS']['PI']]
+        org_ctx = self.get_organization_ctx(ctx)
+        # only one or none so primary is that one
+        if len(pi_list) <= 1 and primary:
+            pi_list[0]['org_ctx'] = org_ctx
+            return pi_list
+        # more than one, get the primary
+        if len(pi_list) > 1 and primary:
+            pi_itr = iter(pi_list)
```
chrisseto: No need to assign this to a variable if you're only using it once

providers/gov/nih/normalizer.py
```diff
@@ -215,14 +281,52 @@ def maybe_org(self, obj):
...
+            while True:
+                try:
+                    pi = next(pi_itr)
+                except StopIteration:
+                    return []
+                if '(contact)' in pi['PI_NAME']:
```
chrisseto: Ah, I think I see. You can replace the entire loop with:
```python
try:
   pi = next(x for x in pi_list if '(contact)' in x['PI_NAME'])
except StopIteration:
    return []
```

providers/gov/nih/normalizer.py
```diff
@@ -215,14 +281,52 @@ def maybe_org(self, obj):
...
+        # more than one, get the non-primary
+        if len(pi_list) > 1 and not primary:
+            non_primary_pi = []
+            for pi in pi_list:
+                if '(contact)' not in pi['PI_NAME']:
+                    pi['org_ctx'] = org_ctx
+                    non_primary_pi.append(pi)
+            return non_primary_pi
+        # only one and not primary
+        return None
```
chrisseto: It might be a bit more clear if you hoist this logic to the beginning of these if blocks

```
if len(x) <= 1:
    if not primary:
        return None
    # logic
    return value

if primary
    # logic
    return value

# logic
return logic
```

PR #543, opened 2016-12-14
[SHARE-499][SHARE-522][Feature] Explicit elastic mappings and better suggestions

aaxelb:
## Purpose
Explicitly define mappings in elasticsearch, rather than depending on dynamic mapping and dynamic templates.

Give better typeahead suggestions in the discover page search facets. Use an approach that relies less on the order of words and allows sorting results by relevance to the typed query, instead of by static weights.

## Changes

- Mapping changes
  - Add `name.autocomplete` fields to agent/tag/source names
    - Use an [edge-ngram analyzer](https://www.elastic.co/guide/en/elasticsearch/guide/current/_index_time_search_as_you_type.html)
    - Remove completion suggester
  - Remove `.raw` multi-field from unanalyzed fields
  - Replace `.raw` with `.exact` on analyzed fields used for filtering (`tags`, `contributors`, `funders`, `identifiers`, `publishers`)
  - Add awards (`lists.funders.awards`)
  - Add `affiliations`, a list of agents the work is related to that don't fit in another field, plus affiliations of its contributors
  - Limit fields included in `_all` (`affiliations`, `contributors`, `description`, `funders`, `hosts`, `identifiers`, `publishers`, `tags`, `title`)
  - Add registration fields from 8136463d3bff8f3841f6a4342a6db513428003a5
    - Also rebase onto that commit, subsuming it.
- Add `ELASTICSEARCH_INDEX_VERSIONS` setting/environment variable
  - Accepts comma-separated list of `vX` versions, e.g. `v1,v2`.
  - Allows running multiple indexes side by side, and selecting which to use in a query parameter to `/search`
    - `?v=2` will search index `share_v2`
    - omitting `v` will search the default index `share`
    - `v` not present in `ELASTICSEARCH_INDEX_VERSIONS` will 400
- Only initialize index and put settings/mappings if  `--es-setup` flag is passed to `runbot elasticsearch`

Companion PR to CenterForOpenScience/ember-share#158

## Ticket
https://openscience.atlassian.net/browse/SHARE-499
https://openscience.atlassian.net/browse/SHARE-522


chrisseto:
Staging this for 2.3. There are a couple other small mapping changes I'd like to make 👍 
We'll pow-wow on it in the next few weeks.

bots/elasticsearch/bot.py
```diff
@@ -25,98 +25,204 @@ def chunk(iterable, size):
...
+                'date_updated': {
+                    'type': 'date',
+                    'format': 'strict_date_optional_time',
+                    'include_in_all': False,
+                },
+                'description': {
                     'type': 'string',
                     'fields': {
-                        'raw': {'type': 'string', 'index': 'not_analyzed'}
+                        'en': {
```
aaxelb: Oh, I should probably just delete the `en` fields. They are included in `_all`, but the values are copied before analysis, so there's no actual benefit unless we explicitly include `title.en`/`description.en` in the query. Exposing `query_string` syntax to the user makes it harder to do fun things with analyzers, unfortunately.

bots/elasticsearch/bot.py
```diff
@@ -25,98 +25,204 @@ def chunk(iterable, size):
...
+                    'include_in_all': False
                 },
-                'types': {
+                'name': {
                     'type': 'string',
-                    'fields': {
-                        'raw': {'type': 'string', 'index': 'not_analyzed'}
-                    }
-                },
+                    'fields': AUTOCOMPLETE_FIELD
```
aaxelb: Sure. On tags and sources, too?

bots/elasticsearch/bot.py
```diff
@@ -150,6 +256,18 @@ def setup(self):
         logger.debug('Waiting for yellow status')
         self.es_client.cluster.health(wait_for_status='yellow')
 
+        logger.info('Putting Elasticsearch settings')
+        self.es_client.indices.close(
+            index=settings.ELASTICSEARCH_INDEX,
+        )
+        self.es_client.indices.put_settings(
```
aaxelb: Ohh, *that's* why I got random "index closed" errors while messing with mappings. 👍

bots/elasticsearch/util.py
```diff
@@ -143,6 +145,29 @@ def fetch_creativework(pks):
...
+                            LEFT JOIN LATERAL (
+                                        SELECT json_agg(json_strip_nulls(json_build_object(
+                                                                            'id', affiliated_agent.id
+                                                                            , 'type', affiliated_agent.type
+                                                                            , 'name', affiliated_agent.name
+                                                                            , 'affiliation_type', affiliation.type
+                                                                        ))) AS affiliations
+                                        FROM share_agentrelation AS affiliation
+                                        JOIN share_agent AS affiliated_agent ON affiliation.related_id = affiliated_agent.id
+                                        WHERE affiliation.subject_id = agent.id
```
aaxelb: Makes sense, I guess.

bots/elasticsearch/util.py
```diff
@@ -143,6 +145,29 @@ def fetch_creativework(pks):
...
+                                        SELECT json_agg(json_strip_nulls(json_build_object(
+                                                                            'id', award.id
+                                                                            , 'name', award.name
+                                                                            , 'description', award.description
+                                                                            , 'uri', award.uri
+                                                                            , 'amount', award.award_amount
+                                                                        ))) AS awards
+                                        FROM share_throughawards AS throughaward
+                                        JOIN share_award AS award ON throughaward.award_id = award.id
+                                        WHERE throughaward.funder_id = agent_relation.id
```
aaxelb: Yeah, the agent here would be a funder, and awards are attached to the `Funder` relation. I'll add `agent_relation.type == 'share.funder'` to make it explicit and probably save time looking for awards on non-funders, come to think of it.

bots/elasticsearch/util.py
```diff
@@ -233,4 +263,8 @@ def fetch_creativework(pks):
                 populate_types(data)
                 data['date'] = (data['date_published'] or data['date_updated'] or data['date_created'])
 
+                for field in ('affiliations', 'contributors', 'funders', 'publishers'):
```
aaxelb: `affiliations` is constructed in python from the agent-work relations that don't fit in another field, plus all the contributors' affiliations. When multiple contributors are affiliated with the same organization, it has dupes, which seemed wrong. Then once I was deduping one list, I figured I may as well do all of them for good measure.

I guess duplicate affiliations aren't a real problem, though... their effect on relevance might even be desirable. More affiliations == more relevant.

bots/elasticsearch/util.py
chrisseto: Do we ever do this query? Wouldn't we want to just index awards on the works they are given to?

aaxelb: It is indexed on the work, in `lists.funders.awards`. All these changes are in `fetch_creativework`, I didn't touch `fetch_agent`.

chrisseto:
Looks good, my only issue is that I cannot read 👍

bots/elasticsearch/bot.py
```diff
@@ -150,6 +257,13 @@ def setup(self):
         logger.debug('Waiting for yellow status')
         self.es_client.cluster.health(wait_for_status='yellow')
 
+        logger.info('Putting Elasticsearch settings')
```
chrisseto: Can this setup be in an `if` block that defaults to false?  That way the existing indexing can be kept alive and well without have to change the jobs or anything. Setting up the new one will just require an extra flag to be set.

chrisseto:
Can you throw into settings `ELASTICSEARCH_ALLOWED_(INDEXES|INDICES)` that loads from the same environ variable splitting on `,`s defaulting to and empty tuple.

Then in the ES search endpoint pop off the query parameter `v`, if `v` exists and is in the aforementioned setting append it with `_` to the ES index.

IE
if `ELASTICSEARCH_ALLOWED_INDEXES` is `v1,v2`
`/search` goes to the `share` index
`/search?v=1` goes to the `share_v1` index
`/search?v=2` goes to the `share_v2` index
`/search?v=3` 400s

I'm not sold on any naming here, just the functionality, please pick something better.

chrisseto:
Last bit (hopefully) Could you also add mappings for the new registration fields? See the commit linked to [this ticket](https://openscience.atlassian.net/browse/SHARE-524)

bots/elasticsearch/bot.py
aaxelb: Sure. Setting up a new index or updating mappings now requires `--es-setup` given to runbot.

aaxelb:
@chrisseto Requests granted. 💝 

In order to add the registration fields to elasticsearch I rebased onto `feature/registries`, then rebased back onto develop, so this branch now subsumes that one.

PR #550, opened 2016-12-22
[SHARE-483][Hotfix] Log disambiguation tasks in CeleryTasks table

aaxelb:
- Rename `MakeJsonPatches` to `DisambiguatorTask`
- Rename `ProviderTask` to `LoggedTask`, and add `AppTask` for tasks that have an associated app (everything but `DisambiguatorTask`)

At first I renamed a couple other things, too (`CeleryProviderTask` -> `CeleryAppTask`, `provider` -> `source`) but the migrations got annoying and it didn't make much difference, anyway. So here it is, migrationless.

## Ticket
https://openscience.atlassian.net/browse/SHARE-483


aaxelb:
Merged into develop (05e9a7c930604afaa8e33359c04e8b60a9d7e84b) and master (84a6dd9bb5bd05dd7ef3e28840e9710d661caadc)

PR #556, opened 2017-01-06
[SHARE-465][Feature] Add endpoint for model types

leb2dg:
## Purpose

Provide a simple way to get all of the creative work types.

## Changes

Add `modeltypes` endpoint that returns the contents of `creative.yaml`.

## Side effects

None

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -94,5 +94,6 @@ urlpatterns = [
     url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
 
     url(r'schema/?$', views.SchemaView.as_view(), name='schema'),
+    url(r'modeltypes/?$', views.ModelTypesView.as_view(), name='modeltypes'),
     *model_schema_patterns
 ] + router.urls
--- a/api/urls.py
+++ b/api/urls.py
@@ -94,6 +94,6 @@ urlpatterns = [
     url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
 
     url(r'schema/?$', views.SchemaView.as_view(), name='schema'),
-    url(r'modeltypes/?$', views.ModelTypesView.as_view(), name='modeltypes'),
+    url(r'schema/creativework/hierarchy/?$', views.ModelTypesView.as_view(), name='modeltypes'),
     *model_schema_patterns
 ] + router.urls
--- a/api/views/schema.py
+++ b/api/views/schema.py
@@ -1,4 +1,7 @@
 import re
+import os
+import yaml
+from collections import OrderedDict
 
 from rest_framework import views
 from rest_framework.response import Response
@@ -6,7 +9,7 @@ from rest_framework.response import Response
 from share import models
 from share.models.validators import JSONLDValidator
 
-__all__ = ('SchemaView', 'ModelSchemaView')
+__all__ = ('SchemaView', 'ModelSchemaView', 'ModelTypesView')
 
 
 INDENT = 4 * ' '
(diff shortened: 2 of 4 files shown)


api/views/share.py
```diff
@@ -127,3 +130,11 @@ def get(self, request, *args, **kwargs):
                 return HttpSmartResponsePermanentRedirect(url)
             return HttpSmartResponseRedirect(url)
         return http.HttpResponseGone()
+
+
+class ModelTypesView(views.APIView):
+    def get(self, request, *args, **kwargs):
+        yaml_file = os.path.abspath('../SHARE/share/models/creative.yaml')
+        with open(yaml_file) as fobj:
+            model_specs = yaml.load(fobj)
```
chrisseto: This should be moved into the class definition so that it will only load once, on start up.

api/views/share.py
```diff
@@ -127,3 +130,11 @@ def get(self, request, *args, **kwargs):
                 return HttpSmartResponsePermanentRedirect(url)
             return HttpSmartResponseRedirect(url)
         return http.HttpResponseGone()
+
+
+class ModelTypesView(views.APIView):
+    def get(self, request, *args, **kwargs):
+        yaml_file = os.path.abspath('../SHARE/share/models/creative.yaml')
```
chrisseto: This path will not work on all machines. Use the module level variable `__file__` to build the path.

api/views/share.py
```diff
@@ -127,3 +130,11 @@ def get(self, request, *args, **kwargs):
...
+        with open(yaml_file) as fobj:
+            model_specs = yaml.load(fobj)
+        return Response(model_specs)
```
chrisseto: Will this be sent back in JSON API format?

api/urls.py
```diff
@@ -94,5 +94,6 @@ def register_url(self, subclass, viewset):
     url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
 
     url(r'schema/?$', views.SchemaView.as_view(), name='schema'),
+    url(r'modeltypes/?$', views.ModelTypesView.as_view(), name='modeltypes'),
```
chrisseto: This should be somewhere a bit more meaningful. Maybe just a generic `hierarchy/<modelname>` or `schema/<modelname>/types|hierarchy`?

api/views/share.py
leb2dg: Yup!

api/urls.py
leb2dg: @chrisseto, are you suggesting that I redesign this entirely to support the hierarchy of all models?

aaxelb: If you do, that'd be cool 👍

chrisseto: @leb2dg Not really, that url pattern leaves room to do so if we'd like to.

api/views/schema.py
```diff
@@ -113,3 +116,24 @@ def get(self, request, *args, **kwargs):
...
     }))
+
+
+class ModelTypesView(views.APIView):
+
+    def __init__(self, *args, **kwargs):
+        super(ModelTypesView, self).__init__(*args, **kwargs)
+        yaml_file = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))) + '/share/models/creative.yaml'
+        with open(yaml_file) as fobj:
+            self.model_specs = self.sort_dict(yaml.load(fobj))
```
chrisseto: This should be done at the class level.

api/views/schema.py
```diff
@@ -113,3 +116,24 @@ def get(self, request, *args, **kwargs):
         'MODEL': model,
         '__doc__': ModelSchemaView.__doc__.format(model.__name__, model.__doc__)
     }))
+
+
+class ModelTypesView(views.APIView):
+
+    def __init__(self, *args, **kwargs):
+        super(ModelTypesView, self).__init__(*args, **kwargs)
+        yaml_file = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__)))) + '/share/models/creative.yaml'
```
chrisseto: Use `os.path.join` over the `+` here. Rather than using multiple `os.path.dirname`s throw `../` into the path.

PR #560, opened 2017-01-11
[SHARE-520][Feature] Encode all ids in API results

leb2dg:
## Purpose

Show encoded ids in API instead of pks.

## Changes

* Override `extract_relationships` from `JSONAPIRenderer`
* Update generated endpoint tests

## Side effects

None

### DIFF
--- a/api/renderers.py
+++ b/api/renderers.py
@@ -50,16 +50,18 @@ class HideNullJSONAPIRenderer(JSONAPIRenderer):
 
         return utils.format_keys(data)
 
-    def encode_ids(relation_data):
-        if relation_data:
-            if not isinstance(relation_data, list):
-                relation_data = [relation_data]
-            for obj in relation_data:
-                obj['id'] = IDObfuscator.encode_id(int(obj['id']), apps.get_model('share', obj['type']))
-        return relation_data
-
     def encode_id(resource_id, resource_type):
         return encoding.force_text(IDObfuscator.encode_id(resource_id, apps.get_model('share', resource_type)))
+
+    @classmethod
+    def encode_ids(cls, relation_data):
+        if relation_data:
+            if isinstance(relation_data, list):
+                for obj in relation_data:
+                    obj['id'] = cls.encode_id(int(obj['id']), obj['type'])
+            else:
+                relation_data['id'] = cls.encode_id(int(relation_data['id']), relation_data['type'])
+        return relation_data
 
     # override ids in relationships from JSONAPIRenderer
     @classmethod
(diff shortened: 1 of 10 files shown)


api/renderers.py
```diff
@@ -46,11 +50,199 @@ def extract_attributes(fields, resource):
...
+                    for position in range(len(serializer_data)):
+                        nested_resource_instance = resource_instance_queryset[position]
+                        nested_resource_instance_type = (
+                            relation_type or
+                            utils.get_resource_type_from_instance(nested_resource_instance)
+                        )
+
+                        relation_data.append(OrderedDict([
+                            ('type', nested_resource_instance_type),
+                            ('id', cls.encode_id(nested_resource_instance.pk, nested_resource_instance_type))
```
chrisseto: If `nested_resource_instance` is just a model instance `IDObfuscator.encode(nested_resource_instance)` will work.

api/renderers.py
```diff
@@ -46,11 +50,199 @@ def extract_attributes(fields, resource):
 
         return utils.format_keys(data)
 
+    def encode_ids(relation_data):
+        if relation_data:
+            if not isinstance(relation_data, list):
```
chrisseto: Won't this transform single relations into lists?

api/renderers.py
```diff
@@ -46,11 +50,199 @@ def extract_attributes(fields, resource):
...
+
+                for related_object in relation_queryset:
+                    relation_data.append(
+                        OrderedDict([('type', relation_type), ('id', cls.encode_id(related_object.pk, relation_type))])
+                    )
+
+                data.update({field_name: {
+                    'links': {
+                        'related': resource.get(field_name)},
+                    'data': cls.encode_ids(relation_data),
```
chrisseto: This looks like it might try to double encode related objects?

leb2dg: 👍  Removed the encoding in the for loop

api/renderers.py
```diff
@@ -46,11 +50,201 @@ def extract_attributes(fields, resource):
...
+                resolved, relation_instance = utils.get_relation_instance(resource_instance, source, field.parent)
+                if not resolved:
+                    continue
+
+                data.update({
+                    field_name: {
+                        'data': (
+                            OrderedDict([
+                                ('type', relation_type),
+                                ('id', cls.encode_id(resource_instance.pk, relation_type))
```
leb2dg: Here I could use `encoding.force_text(IDObfuscator.encode(resource_instance))` instead of `cls.encode_id(resource_instance.pk, relation_type)` but it seems more clear to have all of the encoding use the same method. That way if `IDObfuscator` ever changes, we will only have to change the code in one place.

api/renderers.py
```diff
@@ -46,11 +50,201 @@ def extract_attributes(fields, resource):
...
+
+        return utils.format_keys(data)
+
+    # override top level id from JSONAPIRenderer
     @classmethod
     def build_json_resource_obj(cls, fields, resource, resource_instance, resource_name):
         resource_data = [
             ('type', resource_name),
-            ('id', encoding.force_text(resource_instance.pk) if resource_instance else None),
+            ('id', cls.encode_id(resource_instance.pk, resource_name) if resource_instance else None),
```
leb2dg: Same thing here

PR #565, opened 2017-01-17
[SHARE-511][Feature] Store user favicons as a blob

aaxelb:
Store favicons as a blob in the database.

- add `ShareUser.favicon` field
  - `ImageField` with custom storage to save image as a blob in `FaviconImage` model
  - only editable in admin
- serve favicons at `/favicons/{username}.ico`
  - TODO after merge: cache aggressively
- update favicon urls in various API serializers (used in CenterForOpenScience/ember-share#168)
- provider migrations to save favicon 
  - move `favicon.ico` to app root, out of sight of staticfiles

https://openscience.atlassian.net/browse/SHARE-511


bots/autocurate/person/migrations/0002_favicon.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
chrisseto: The bot migrations can be removed.

share/robot.py
```diff
@@ -166,3 +206,21 @@ def __call__(self, apps, schema_editor):
...
+
+class RobotFaviconMigration(AbstractRobotMigration):
+
+    def __call__(self, apps, schema_editor):
+        user = self.config.user
+        try:
+            with open(os.path.join(self.config.path, 'favicon.ico'), 'rb') as f:
+                user.favicon.save(user.username, File(f))
+        except OSError:
+            pass
```
chrisseto: This exception shouldn't be caught.  Anything that has this migration should have a favicon as well.

share/models/core.py
```diff
@@ -75,6 +81,37 @@ def create_robot_user(self, username, robot, long_title='', home_page=''):
...
+
+@deconstructible
+class FaviconStorage(Storage):
+    def _open(self, name, mode='rb'):
+        assert mode == 'rb'
+        favicon = FaviconImage.objects.get(user__username=name)
+        return ContentFile(favicon.image)
+
+    def _save(self, name, content):
+        user = ShareUser.objects.get(username=name)
```
chrisseto: Would a join in `update_or_create` work?

share/models/core.py
```diff
@@ -75,6 +81,37 @@ def create_robot_user(self, username, robot, long_title='', home_page=''):
         return user
 
 
+class FaviconImage(models.Model):
+    user = models.OneToOneField('ShareUser', on_delete=DATABASE_CASCADE, primary_key=True)
```
chrisseto: Let django handle the primary key. Things have a tendency to get weird when you don't.

share/graphql/base.py
```diff
@@ -26,7 +26,7 @@ def resolve_title(cls, instance, context, request, info):
 
     @classmethod
     def resolve_favicon(cls, instance, context, request, info):
-        return '/static/{}/img/favicon.ico'.format(instance.robot.replace('providers.', '', 1))
```
chrisseto: Just for sanity's sake, can you use django's reverse URL look up here?

requirements.txt
```diff
@@ -55,3 +55,4 @@ newrelic==2.68.0.50
 raven==5.23.0
 graphene==1.0.2
 graphene-django==1.0
+pillow==4.0.0
```
chrisseto: Is pillow actually being used for anything?

aaxelb: It's required by `ImageField`, to validate that uploaded files are images.

share/graphql/base.py
aaxelb: Whoops, yes.

share/models/core.py
aaxelb: I'd tried `user__username=name` but it complained that `user_id` was missing. Apparently it's not smart enough for that.

share/models/core.py
aaxelb: 👍

share/robot.py
aaxelb: Makes sense. I'll add the missing favicons, then, too.

bots/autocurate/person/migrations/0002_favicon.py
aaxelb: I'll only add to apps with `ProviderAppConfig`s.

aaxelb:
@chrisseto changes made

api/views/share.py
```diff
@@ -107,6 +108,15 @@ def get(self, request, *args, **kwargs):
         return Response(ser.data)
 
 
+def user_favicon_view(request, username):
+    user = ShareUser.objects.get(username=username)
```
chrisseto: Looks like this guy 500's rather than 404's if it fails

This should also only respond to GETs, just for semantic correctness.

aaxelb: Oh, derp. Didn't think about the *user* not existing. Fixed.

PR #566, opened 2017-01-18
[SHARE-488][Feature] Add SSOAR harvester

leb2dg:


chrisseto:
@leb2dg Could you update this to work with the favicons change?

PR #567, opened 2017-01-23
[SHARE-522][Feature] Gevent elasticsearch tasks

chrisseto:


project/settings.py
```diff
@@ -471,39 +471,41 @@
...
-HIGH_PRI_MODULES = {
-    'share.tasks.BotTask',
-    'bots.elasticsearch',
+QUEUES = {
+    'DEFAULT': {
+        'name': 'celery',
+        'priority': 0,
+        'modules': set(),
+    },
+    'GEVENT': {
```
aaxelb: Should this queue only exist depending on the `GEVENT` env variable?  Or does it make no difference

aaxelb:
Nice, much cleaner than before 👍

share/celery.py
```diff
@@ -5,13 +5,10 @@ def match_by_module(task_path):
...
-            return settings.LOW_QUEUE
-        if task_subpath in settings.MED_PRI_MODULES:
-            return settings.MED_QUEUE
-        if task_subpath in settings.HIGH_PRI_MODULES:
-            return settings.HIGH_QUEUE
-    return settings.DEFAULT_QUEUE
+        for v in settings.QUEUES.values():
+            if task_subpath in v['modules']:
+                return v['name']
+    return v['DEFAULT']['name']
```
aaxelb: Missed this before... should be `settings.QUEUES` instead of `v`

PR #575, opened 2017-01-27
[SHARE-514][Feature] Add repository as a type

leb2dg:
## Purpose
Repository is a type needed to push UCSD data and not all of their awards have URIs.

## Changes
Add `Repository` type
Do not require URI on award
Add migration
Don't disambiguate department on name

## Side effects
None


chrisseto:
Added with https://github.com/CenterForOpenScience/SHARE/releases/tag/2.2.5

PR #576, opened 2017-01-27
[SHARE-538][SHARE-539][Feature] Merging objects

aaxelb:
- [x] Emit merge nodes when disambiguation returns multiple objects
    e.g. `{'@id': '2D1DE-ADB-EEF', '@type': 'creativework', 'same_as': {'@id': '2D1DE-ADB-EE5', '@type': 'creativework'}}`
- [x] Handle `same_as`
  - [x] Set the target's `same_as`, `same_as_version`
  - [x] Update all foreign keys to the target to point to the same as `same_as`
  - [x] Handle conflicts -- when simply updating a foreign key violates a unique constraint, merge the conflicting things together, setting `same_as` on the old.
  - [x] Merge scalar fields on the objects. Take the value from the more recently modified object, ignoring blank/null values.
- [x] Remove objects with non-null `same_as` from elastic index
- [x] Resolve API requests for merged objects to the `same_as` object instead (both REST and graphql)
- [x] Add management command to merge objects
    e.g. `python manage.py merge 2D1DE-ADB-EEF 2D1DE-ADB-EE7 --into 2D1DE-ADB-EE5`
    TODO (in a large, eventual refactor): "undo"
- [x] Thoroughly test

https://openscience.atlassian.net/browse/SHARE-538
https://openscience.atlassian.net/browse/SHARE-539


api/views/share.py
```diff
@@ -92,6 +92,8 @@ def get_object(self):
 
         filter_kwargs = {self.lookup_field: decoded_pk}
         obj = get_object_or_404(queryset, **filter_kwargs)
+        while obj.same_as:
```
chrisseto: Not an issue now, but this would be better done as a recursive query.

share/models/base.py
```diff
@@ -169,6 +170,9 @@ class ShareObject(models.Model, metaclass=ShareObjectMeta):
     class Meta:
         abstract = True
 
+    def __repr__(self):
```
chrisseto: ❤️

share/models/change.py
```diff
@@ -272,3 +235,61 @@ def _resolve_change(self):
             else:
                 change[k] = v
         return change
+
+    def _merge_objects(self, from_obj, into_obj):
+        from share.models.base import ShareObject
+
+        # Avoid same_as chains
+        while into_obj.same_as:
```
chrisseto: See API comment

share/models/change.py
```diff
@@ -272,3 +235,61 @@ def _resolve_change(self):
...
+            into_obj = into_obj.same_as
+        for obj in from_obj._meta.concrete_model.objects.filter(same_as=from_obj.id):
```
chrisseto: Please make this into a single query.

share/models/change.py
```diff
@@ -272,3 +235,61 @@ def _resolve_change(self):
...
+                m = re.search('Key \(([^)]+)\)=\(([^)]+)\) already exists.', e.args[0])
+                if m:
+                    keys = dict(zip(m.group(1).split(', '), m.group(2).split(', ')))
+                    conflicting_obj = obj._meta.model.objects.get(**keys)
+                    obj.refresh_from_db()
+                    self._merge_objects(obj, conflicting_obj)
+                else:
+                    raise e
+
+    def _merge_scalar_field(self, from_obj, into_obj, field):
```
chrisseto: Not necessary right now, but this would be a good place to throw in a fuzzy match on strings.
https://en.wikipedia.org/wiki/Approximate_string_matching

share/disambiguation.py
```diff
@@ -127,22 +120,19 @@ def _instance_for_node(self, node):
 
         for q in constrain:
```
aaxelb: I'm not sure... if there's a person with the same name as an institution, the loose constraints will match both, I think? And we don't want to merge them together. There's likely a simpler way to do this, but theres [a ticket](https://openscience.atlassian.net/browse/SHARE-549) for that.

api/views/share.py
aaxelb: Good call. I added a `get_canonical` method to ShareObjectManager, which uses a recursive query to follow same_as to the next unmerged object, and detect same_as cycles (not that they could ever happen...)

share/models/change.py
aaxelb: I'll make a ticket for that.

aaxelb:
@chrisseto Followed your suggestions, and noticed several flaws/omissions that I fixed. Ready for round 2...

share/models/change.py
```diff
@@ -89,7 +93,7 @@ def accept(self, save=True):
         ret = []
         with transaction.atomic():
             self._changes_cache = list(self.changes.all())
-            for c in self._changes_cache:
+            for c in sorted(self._changes_cache, key=lambda c: c.type):
```
chrisseto: This feels a bit scary. The order of the changes does matter IIRC. If it keeps the order within the types, you're probably good.

share/models/change.py
```diff
@@ -271,3 +242,66 @@ def _resolve_change(self):
             else:
                 change[k] = v
         return change
+
+    def _merge_objects(self, from_obj, into_obj):
+        from share.models import ShareObject
+
+        # TODO less hacky way of expressing what can be merged
```
chrisseto: Please?

Move this to either a class level or global scope.

share/models/change.py
```diff
@@ -271,3 +242,66 @@ def _resolve_change(self):
...
+        MERGEABLE_MODELS = {'AbstractCreativeWork', 'AbstractAgent', 'Award', 'AbstractWorkRelation', 'AbstractAgentRelation', 'AbstractAgentWorkRelation', 'ThroughTags', 'ThroughSubjects', 'ThroughContributor', 'ThroughAwards'}
+        concrete_model = from_obj._meta.concrete_model
+        if concrete_model.__name__ not in MERGEABLE_MODELS:
+            raise InvalidMergeError('Invalid model for merging: %s' % concrete_model)
```
chrisseto: This is an odd inconsistency, but use new style formatting on exceptions.

The only case for old style format is log messages. There may be additional compute time required to format the string that can be avoided if the log is too high/low of a level.

share/models/change.py
```diff
@@ -271,3 +242,66 @@ def _resolve_change(self):
...
+
+        from_obj.change = self
+        from_obj.same_as = into_obj
+        from_obj.same_as_version = into_obj.version
+        from_obj.save()
+
+        return into_obj
+
+    def _merge_fk_field(self, from_obj, into_obj, field):
+        for obj in field.model.objects.filter(**{field.name: from_obj.id}):
```
chrisseto: This looks like it would be handled better with an UPSERT?

share/tasks.py
```diff
@@ -208,7 +208,7 @@ def do_run(self, *args, **kwargs):
                 cg = ChangeGraph(self.normalized.data['@graph'], namespace=self.normalized.source.username)
                 cg.process()
                 cs = ChangeSet.objects.from_graph(cg, self.normalized.id)
-                if cs and (self.source.is_robot or self.source.is_trusted):
+                if cs and (self.source.is_robot or self.source.is_trusted or self.source.username == settings.APPLICATION_USERNAME):
```
chrisseto: `is_superuser` would be a better flag to check here.

tests/share/normalize/factories.py
```diff
@@ -363,7 +363,7 @@ def _parse(self, *args, **kwargs):
 
 
 def _params(seed=None, id=None, type=None, **kwargs):
-    string_id = '_:_' + str(id)
+    string_id = id if isinstance(id, str) else '_:_' + str(id)
```
chrisseto: What is this actually used for?

share/change.py
```diff
@@ -298,7 +298,7 @@ def __init__(self, graph, id, type, attrs, namespace=None):
         self.namespace = namespace
 
         if not self.is_blank:
-            self.instance = IDObfuscator.load(self.id, None)
+            self.instance = IDObfuscator.resolve(self.id)
```
aaxelb: I changed `resolve` to use `get_canonical`, so it'll follow the `same_as` chain. I didn't notice that changed its behavior to return None instead of raising NotFoundError... How about I change `load` to use `get_canonical`, and get rid of `resolve`?

chrisseto: Hmm... one way or another I don't think both should exist. However whatever name we go with should also take a `canonical=False` (default to `True`) that will just load the non-canonical record.

I'm partial to `load` just because its fewer characters that I have to type. 🍆

aaxelb: 👍

share/disambiguation.py
```diff
@@ -64,21 +65,17 @@ def _disambiguate(self, change_graph, find_instances):
...
-                    #     logger.warning('Found multiple matches for %s, but only one of type %s, fortunately.', n, n.model)
-                    #     instance = same_type.pop()
-                    if instance:
+                    instances = self._instances_for_node(n)
+                    if instances:
+                        if n.instance:
+                            instances = list(set(instances + [n.instance]))
+                        if len(instances) == 1:
+                            n.instance = instances[0]
+                            logger.debug('Disambiguated %s to %s', n, repr(n.instance))
```
aaxelb: Oh, neat.

share/models/change.py
aaxelb: Makes sense. I definitely wrote these like they were log messages and made the minimal possible change when I realized that was wrong.

share/tasks.py
aaxelb: That's what I did at first, but `system` isn't a superuser. Should the `merge` command submit its changes as `admin`?

tests/share/normalize/factories.py
aaxelb: For submitting changes with a non-blank SHARE ID (like to add an identifier to a specific work, triggering a merge). Something like `_:_461C1-98C-F8E` doesn't work.

share/models/change.py
aaxelb: Oh, whoops. At first it was a set of the actual models, which couldn't be at global scope, but strings totally can.

share/models/change.py
aaxelb: `sorted` is a stable sort, so this shouldn't be a problem. But now that you mention it, it would be better to but merge nodes first while sorting the ChangeGraph, so the order is always consistent.

chrisseto: Ah! stable is the word that I was looking for. Can you just throw a comment in that `sorted` is stable so my hair doesn't stand on end whenever I read that?

I know I read that it is somewhere but it wasn't in the docstring:
```
Help on built-in function sorted in module builtins:

sorted(iterable, key=None, reverse=False)
    Return a new list containing all items from the iterable in ascending order.

    A custom key function can be supplied to customise the sort order, and the
    reverse flag can be set to request the result in descending order.
```

But:
```
Timsort is a hybrid stable sorting algorithm... Timsort has been Python's standard sorting algorithm since version 2.3.
```

share/tasks.py
chrisseto: Hmmm, maybe we should make system a superuser?

share/models/change.py
aaxelb: There's no case where we want an insert, though... if there's no conflict, it's just an update. If there is a conflict, there are two updates (one to merge field values into the winner of the conflict, one to update the other's `same_as`). I made it recurse into `_merge_objects` so relation fields would be merged by the same logic. And just in case, say, we're merging an Agent with duplicate/conflicting Funder relations with associated Awards...

share/models/change.py
aaxelb: Oh, I already changed ChangeGraph to sort merge nodes first, which I think is much better.

(Looks like the `sorted` docs have been unclear [since 2009](https://stackoverflow.com/questions/1915376/is-pythons-sorted-function-guaranteed-to-be-stable))

share/models/change.py
chrisseto: I mostly really don't enjoy the parsing of an `IntegrityError`...
It's a little bit too late in the day, but I think some fancy sql could handle this pretty well.

share/tasks.py
aaxelb: That seems to make sense... Who do we trust more than the system itself?

I'm not sure the best way to do it, though. Would that be a migration?

share/models/change.py
aaxelb: Yeah, fair, that is super ugly. I'll revisit it tomorrow, and we should probably discuss.

share/disambiguation.py
```diff
@@ -125,24 +122,22 @@ def _instance_for_node(self, node):
             constrain.append(Q(type__in=node.model.get_types()))
             constrain.append(Q(type=node.model._typedmodels_type))
 
+        # HACK
+        unmerged_query = Q(same_as__isnull=True) if hasattr(concrete_model, 'same_as') else Q()
+
         for q in constrain:
```
aaxelb: What if we use the existing `constrain_types` logic for agents, instead of this `constrain` list?
Disambiguating an `institution` would match `agent`, `organization`, or `institution`, so it'll merge in any generic agents with the same name. Disambiguating an `organization` would match `agent`, `organization`, `institution`, `consortium`, or `department`, which could potentially merge things that shouldn't be, but only with the same name/identifier/cited_as, and never people.

aaxelb:
@chrisseto Main changes since your last review, in addition to what you explicitly asked for:
- simplified disambiguation logic (in `_instance_for_node`), since returning a list is okay now
  - `constrain_types=True` on agents, so we don't merge people and organizations together
- look at `unique_together` instead of parsing exception messages while merging
- sort merge nodes first in the ChangeGraph, don't reorder them later

share/disambiguation.py
```diff
@@ -115,34 +112,15 @@ def _instance_for_node(self, node):
...
-            if all('__' in str(query) for query in queries):
-                logger.warning('Multiple %ss returned for %s (The any query) bailing', concrete_model, queries)
-                break
-
-        logger.error('Could not disambiguate %s. Too many results found from %s %s', node.model, all_query, queries)
-        raise NotImplementedError('Multiple {0}s found'.format(node.model))
+        unmerged_query = Q(same_as__isnull=True) if hasattr(concrete_model, 'same_as') else Q()
+
+        sql, params = zip(*[concrete_model.objects.filter(unmerged_query & all_query & query).query.sql_with_params() for query in queries or [Q()]])
+        return list(concrete_model.objects.raw(' UNION '.join('({})'.format(s) for s in sql) + ';', sum(params, ())))
```
chrisseto: Can you throw a check/limit in here to avoid bad queries or loading way to much.

If we some how get 500 results we don't want to continue with that. Should probably just crash or skip it.
Loading all 500 will also take a non-trivial amount of time as well.

aaxelb: Gooood call. I feel like the limit could be quite low, I'd be concerned about merging 10 or even 5 things together.

PR #578, opened 2017-01-30
[SHARE-514][Hotfix] Add null=True to uri on award model

leb2dg:


leb2dg:
Included in https://github.com/CenterForOpenScience/SHARE/commit/8befea8fce2752cab98b282e8815e721dec1b262

PR #581, opened 2017-02-02
[SHARE-529][SHARE-590][Feature] Use cursor, not offset, in Crossref harvester.

aaxelb:
So it doesn't break when harvesting more than 10k records.

Also, keep only a list of RawData IDs when harvesting, not all the RawData instances.

https://openscience.atlassian.net/browse/SHARE-529
https://openscience.atlassian.net/browse/SHARE-590

### DIFF
--- a/providers/org/crossref/harvester.py
+++ b/providers/org/crossref/harvester.py
@@ -15,26 +15,21 @@ class CrossRefHarvester(Harvester):
                 start_date.isoformat(),
                 end_date.isoformat()
             ),
-            'rows': 1000
-        }).url)
+            'rows': 1000,
+        }))
 
-    def fetch_records(self, url):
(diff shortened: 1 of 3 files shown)


chrisseto:
@aaxelb any concern with eating too much memory here? All the docs will end up being held in memory.

aaxelb:
@chrisseto Yes, [I made a ticket for it](https://openscience.atlassian.net/browse/SHARE-590), though I wasn't sure if it was a big enough issue to fix right now. I could definitely add that to this PR, though.

chrisseto:
@aaxelb Could you go ahead and throw that in here? Running out of memory will affect more just this task.

aaxelb:
@chrisseto 👍

PR #584, opened 2017-02-07
[SHARE-497][SHARE-526][Task] Add a status/version endpoint to the API

chrisseto:
* In DEBUG mode, if VERSION is not set it will be pulled from a git
    repo
  * The Dockerfile now accepts GIT_TAG which will be passed in as
    VERSION
  * Also removes "Notify" from browsable API page title

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -52,7 +52,9 @@ COPY ./ /code/
 
 RUN python manage.py collectstatic --noinput
 
+ARG GIT_TAG=
 ARG GIT_COMMIT=
+ENV VERSION ${GIT_TAG}
 ENV GIT_COMMIT ${GIT_COMMIT}
 
 CMD ["python", "manage.py", "--help"]
--- a/api/urls.py
+++ b/api/urls.py
@@ -91,6 +91,7 @@ model_schema_patterns = [
 ]
 
 urlpatterns = [
+    url(r'status/?', views.ServerStatusView.as_view(), name='status'),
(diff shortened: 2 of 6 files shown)


Dockerfile
```diff
@@ -52,7 +52,9 @@ COPY ./ /code/
 
 RUN python manage.py collectstatic --noinput
 
+ARG GIT_TAG=
 ARG GIT_COMMIT=
```
aaxelb: Is there a reason for the separate GIT_TAG and GIT_COMMIT? It [looks like](https://github.com/CenterForOpenScience/SHARE/blob/develop/project/settings.py#L390) GIT_COMMIT is only used for `release` in sentry issues, and it looks like it's null. Seems like they could be the same thing?

chrisseto: Mostly readability:

```
# git rev-parse HEAD
aeb92b8c98b9b2165b66f4ba2f866636743fec26

# git describe
2.3.1-16-gaeb92b8
```

The purpose of the endpoint is mostly to check that the latest release is actually whats released.

I don't really have strong feelings one way or the other. Do you have a preference?

(And I copied it from fakecas.)

aaxelb: `git describe` is definitely better than the full commit hash. I don't have any feelings about GIT_COMMIT, it just seemed potentially redundant, and if it's only used for display (or not actually given a value?) could be replaced with GIT_TAG. :man_shrugging:

PR #589, opened 2017-02-15
[SHARE-604][Fix] Don't accept changes from the past

chrisseto:


tests/share/disambiguation/test_creativework.py
```diff
@@ -174,3 +174,88 @@ def test_no_merge_on_blank_value(self, Graph):
...
+        assert older_graph.nodes[0].change == {'title': 'Not Previously Changed'}
+
+    def test_no_timetraveling_many(self, Graph):
+        oldest_graph = ChangeGraph(Graph(
+            Publication(
+                id=1,
+                sparse=True,
+                is_deleted=True,
+                title='The first title',
+                description='The first language',
```
leb2dg: Should this be 'The first description'?

share/change.py
```diff
@@ -276,8 +276,27 @@ def change(self):
...
+
+                # Go back until we find a change that is older than us
+                for version in self.instance.versions.select_related('change').all():
+                    if not version.date_updated or date_updated > version.date_updated:
+                        break
+                    ignore_attrs.update(version.change.change.keys())
+
         attrs = {}
         for k, v in self.attrs.items():
+            if k in ignore_attrs:
```
leb2dg: Will `ignore_attrs` only contain the keys for the change that resulted in the current instance (not the new change) and the change for the previous version?

tests/share/disambiguation/test_creativework.py
chrisseto: Yes.

share/change.py
chrisseto: It will contain the keys of all changes that have a `date_modified` > the one in the current change.

leb2dg: Ah got it 👍  (`date_updated` not `date_modified`?)

leb2dg:
Looks good to me 💵

PR #595, opened 2017-02-24
Update apps.py

Stevenholloway:
We would like to change the provider name in the "Sources" list to "Scholarly Commons @ JMU"  We would also prefer to have all sets harvested, so I have eliminated the "approved sets" list from the harvester.  Thanks for coding this harvester!  --Steven Holloway, Head of Metadata Strategies, James Madison University


chrisseto:
Thank you @Stevenholloway!

Your changes will be going into release 2.5.0. We're hoping to roll it out some time next week.

Stevenholloway:
Thanks, guys!

PR #596, opened 2017-02-24
[SHARE-558][SHARE-627][SHARE-628][SHARE-630][Feature] Transformers and SourceConfigs

aaxelb:
In broad strokes, this PR:
- removes all the `providers.*` apps
  - source configuration moved to the database (initialized from yaml in `share/sources`) as `Source` and `SourceConfig` models
  - harvesters and transformers (née normalizers) moved to `share.harvesters` and `share.transformers`, where they can easily be used by any source
  - replace 'provider' with 'source' almost everywhere
- deals with the cascading consequences

### Notes for reviewers
- I'm sorry.
  - Most of the harvesters and transformers are barely changed. `git blame` knows this, but the diff still sees it as a delete/add.
- Please focus especially on:
  - `loadsources` management command
    - Adds Harvesters and Transformers according to the entry points in `setup.py` (if needed)
    - Adds or updates Sources, ShareUsers, and SourceConfigs according to `source.yaml` files in `share/sources`
    - Adds PeriodicTasks for each SourceConfig (if needed)
  - migrations
    - migration (a) adds the new tables/columns and deletes things unnecessary for (b)
    - migration (b) runs `loadsources` (is that a bad idea?) and creates SUIDs for all existing raw data
    - migration (c) drops the fields on rawdata that SUIDs replace


api/views/workflow.py
```diff
@@ -32,11 +31,11 @@ def get_queryset(self):
...
 
 
-class ProviderViewSet(viewsets.ReadOnlyModelViewSet):
-    serializer_class = ProviderSerializer
+class SourceViewSet(viewsets.ReadOnlyModelViewSet):
+    serializer_class = SourceSerializer
 
     def get_queryset(self):
-        queryset = ShareUser.objects.exclude(robot='').exclude(long_title='')
+        queryset = Source.objects.exclude(user__robot='').exclude(long_title='')
```
chrisseto: Any reason this shouldn't just be `Source.objects.all()`?

api/views/workflow.py
```diff
@@ -228,3 +226,23 @@ def post(self, request, *args, **kwargs):
             return Response({'task_id': async_result.id}, status=status.HTTP_202_ACCEPTED)
         else:
             return Response({'errors': serializer.errors, 'data': prelim_data}, status=status.HTTP_400_BAD_REQUEST)
+
+    def _get_source_config(self, user):
+        config_label = '{}.v1_push'.format(user.username)
+        try:
+            return SourceConfig.objects.get(label=config_label)
```
chrisseto: I think we should just return a 400 or 403 if the config is not found? Use of older APIs should be explicitly allowed by us.

bots/elasticsearch/util.py
```diff
@@ -62,7 +62,8 @@ def fetch_agent(pks):
                 LEFT JOIN LATERAL (
                             SELECT array_agg(source.long_title) AS sources
```
chrisseto: Maybe `DISTINCT`? I'm not sure if it actually matters to ES.

api/views/workflow.py
aaxelb: This excludes push sources, like before. But I now see there's a ticket to display push sources, so I could change this to `all()` and add yet another ticket to the title.

api/views/workflow.py
aaxelb: Yeah, that makes sense. Is there already a way to authorize people to use v1 (where the source and config should be created)?

bots/elasticsearch/util.py
aaxelb: It doesn't particularly matter for ES, especially since we filter sources by exact match, but may as well 👍

aaxelb:
Subsumed by #601

PR #598, opened 2017-03-02
[SHARE-538][Feature][Fix] Fix some merging errors.

aaxelb:
Make disambiguation and merging more resilient when accepting changesets that (a) trigger a merge and (b) include objects related to the objects to be merged. For example:
- Given a work that has a contributor without an identifier: Push the work's payload a second time with an identifier on its contributor that causes the agent to merge.
- Given two works that have the same tag: Push one of the works a second time (including its tags) with a new identifier that causes the works to merge.

### Changes
- Don't stop disambiguating a node after it's matched one instance. It could match something else on a subsequent pass.
- When accepting a change, follow `target.same_as` in case it was merged into something else earlier in the changeset.
- Don't emit merge nodes for objects that will be implicitly merged by merging a work, agent, or award.

### Side effects
Disambiguating will be somewhat slower. It doesn't seem too bad locally, but something to watch for on a larger db.

### DIFF
--- a/share/disambiguation.py
+++ b/share/disambiguation.py
@@ -36,6 +36,7 @@ class GraphDisambiguator:
 
     def _disambiguate(self, change_graph, find_instances):
         changed = True
+        finished_nodes = set()
         # Sort by type and id as well to get consitent sorting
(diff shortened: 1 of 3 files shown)


chrisseto:
@aaxelb Broken tests

aaxelb:
Arg. Fixed locally, travis pending. @chrisseto

PR #599, opened 2017-03-14
[SHARE-653][Improvement] Display trusted users and robots on sources page

leb2dg:
## Purpose
Display trusted users as sources on the 'Sources' page.

## Changes
* Update the autocomplete on the sources page (requires the reindexing)
* Switch all instances of filtering by robot='' to favicon=''

## Side effects
The logic for which sources appear in the sources list (only on the sources page) changed. It is hard to tell which ones will appear on the production data set since it is based on each ShareUser's settings.

These updates require the ShareUsers to be reindexed in elasticsearch.

### DIFF
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -36,7 +36,7 @@ class ProviderViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = ProviderSerializer
 
     def get_queryset(self):
-        queryset = ShareUser.objects.exclude(robot='').exclude(long_title='')
+        queryset = ShareUser.objects.exclude(long_title='').exclude(favicon='')
         sort = self.request.query_params.get("sort")
         if sort:
             return queryset.order_by(sort)
--- a/bots/elasticsearch/tasks.py
+++ b/bots/elasticsearch/tasks.py
@@ -121,13 +121,13 @@ class IndexSourceTask(AppTask):
     def bulk_stream(self, es_index):
         ShareUser = apps.get_model('share.ShareUser')
         opts = {'_index': es_index, '_type': 'sources'}
-        for source in ShareUser.objects.exclude(robot='').exclude(long_title='').all():
-            yield {'_op_type': 'index', '_id': source.robot, **self.serialize(source), **opts}
+        for source in ShareUser.objects.exclude(long_title='').exclude(favicon='').all():
+            yield {'_op_type': 'index', '_id': source.username, **self.serialize(source), **opts}
 
     def serialize(self, source):
         return {
-            'id': source.robot,
+            'id': source.username,
             'type': 'source',
             'name': safe_substr(source.long_title),
-            'short_name': safe_substr(source.robot)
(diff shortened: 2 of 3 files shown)


api/views/workflow.py
```diff
@@ -36,7 +36,7 @@ class ProviderViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = ProviderSerializer
 
     def get_queryset(self):
-        queryset = ShareUser.objects.exclude(robot='').exclude(long_title='')
+        queryset = ShareUser.objects.exclude(long_title='').exclude(favicon='')
```
aaxelb: Should this also exclude `trusted=False` for push users?

bots/elasticsearch/tasks.py
```diff
@@ -121,13 +121,13 @@ def do_run(self, es_url=None, es_index=None):
     def bulk_stream(self, es_index):
         ShareUser = apps.get_model('share.ShareUser')
         opts = {'_index': es_index, '_type': 'sources'}
-        for source in ShareUser.objects.exclude(robot='').exclude(long_title='').all():
-            yield {'_op_type': 'index', '_id': source.robot, **self.serialize(source), **opts}
+        for source in ShareUser.objects.exclude(long_title='').exclude(favicon='').all():
```
aaxelb: Ditto, `trusted`?

share/graphql/query.py
```diff
@@ -41,7 +41,7 @@ def resolve_me(self, args, context, info):
     @graphene.resolve_only_args
     def resolve_sources(self, limit=25, offset=0):
         # TODO Long title will need an index
-        return models.ShareUser.objects.exclude(robot='').exclude(long_title='').order_by('long_title')[offset:offset + limit]
+        return models.ShareUser.objects.exclude(long_title='').exclude(favicon='').order_by('long_title')[offset:offset + limit]
```
aaxelb: Tritto?

api/views/workflow.py
leb2dg: Robots have `is_trusted=False` but theoretically push users who are not trusted should not have favicons. Not filtering out `is_trusted=False` also leaves the flexibility of accepting changes from a user who is not trusted and giving them a favicon so that they will show up as a source. Any flaws in that logic?

leb2dg: Oh plus we don't want all of the OSF push users to show up (forgot about the main reason).

aaxelb: That makes sense! I may have been confused by the title/description of this PR. It's probably close enough to intuitive that in order to be prominently displayed, you need a favicon. I'm good with that.

Is there more than one OSF push user? I thought the other OSF users are just logging into share, and are not trusted.

leb2dg: Ah fair enough, probably should have been more clear with the title. 

You're right about the OSF push user thing, I was looking at the remnants of when we were harvesting the preprint services and didn't actually check their permissions.

PR #605, opened 2017-03-27
[Fix] Various HarvestLog fixes/cleanup (See full msg)

chrisseto:
* Don't clean dates for a task that has already started
    * Add colors for all log status
    * Add a bulk retry/restart button


share/admin.py
```diff
@@ -227,14 +241,22 @@ def end_date_(self, obj):
         return obj.end_date.date().isoformat()
 
     def status_(self, obj):
-        if obj.status == HarvestLog.STATUS.created and (timezone.now() - obj.date_modified) > datetime.timedelta(minutes=60):
-            return '<span style="font-weight: bold;">Lost</span>'
+        if obj.status == HarvestLog.STATUS.created and (timezone.now() - obj.date_modified) > datetime.timedelta(days=1, hours=6):
+            return format_html('<span style="font-weight: bold;">Lost</span>')
```
leb2dg: Should `lost` be another status? What would cause something to be "lost"?

share/admin.py
```diff
@@ -227,14 +241,22 @@ def end_date_(self, obj):
...
         return format_html(
             '<span style="font-weight: bold; color: {}">{}</span>',
-            ['blue', 'cyan', 'red', 'green', 'goldenrod', 'orange'][obj.status],
+            self.STATUS_COLORS[obj.status_identifier],
             HarvestLog.STATUS[obj.status].title(),
         )
 
+    def restart_tasks(self, request, queryset):
+        for log in queryset.select_related('source_config'):
+            HarvesterTask().apply_async((1, log.source_config.label), {
```
aaxelb: Should this have `superfluous`? Or do you see this button just for restarting failed/lost/etc. tasks

share/admin.py
```diff
@@ -213,6 +214,19 @@ class HarvestLogAdmin(admin.ModelAdmin):
...
     readonly_fields = ('harvest_log_actions',)
+    actions = ('restart_tasks', )
+
+    STATUS_COLORS = {
+        'created': 'blue',
+        'started': 'cyan',
+        'failed': 'red',
+        'succeeded': 'green',
+        'rescheduled': 'goldenrod',
+        'defunct': 'purple',
```
aaxelb: I maybe should have thought about this before, but how does something become defunct?

share/admin.py
chrisseto: `lost` is something that hasn't had its status changed in a 1.25 days. IE the task was created but was never run. I don't think it qualifies as a status, there is nothing that triggers the lost state. It's just an assumption/something to quickly draw the eye.

share/models/logs.py
```diff
@@ -107,6 +108,10 @@ class Meta:
         abstract = True
         ordering = ('-date_modified', )
 
+    @property
+    def status_identifier(self):
+        return next(k for k, v in type(self).STATUS._identifier_map.items() if v == self.status)
```
chrisseto: I'm mostly just using it for the color map in the admin. Seemed a bit nicer, should that just be changed over as well?

share/admin.py
chrisseto: Ah, I'll add some comments next to the status definitions.

defunct would be if a task is started with an older version of a harvester/transformer. And I also just realized that we default to the latest version of everything so this case will never be hit... Whoops. Will remove.

share/models/logs.py
aaxelb: Seems like the color map should use `AbstractBaseLog.STATUS.created`, etc as keys. A bit more verbose but less brittle.

chrisseto: 👍

share/admin.py
chrisseto: Yup, this is just a bulk restart 👍

leb2dg:
@chrisseto, there is a failing "behave" test related to the defunct change.

share/models/logs.py
```diff
@@ -76,9 +108,9 @@ class AbstractBaseLog(models.Model):
         (2, 'failed', _('Failed')),
         (3, 'succeeded', _('Succeeded')),
         (4, 'rescheduled', _('Rescheduled')),
-        (5, 'defunct', _('Defunct')),
```
leb2dg: So `1, 2, 3, 4, 6, 7, 8`... no `5`? Is that necessary?

chrisseto: Added a comment to address this.

PR #609, opened 2017-03-28
[SHARE-641][Feature] SHARE OAI-PMH endpoint

aaxelb:
https://openscience.atlassian.net/browse/SHARE-641


share/oaipmh/views.py
```diff
@@ -0,0 +1,256 @@
...
+            prefix = kwargs['metadataPrefix']
+            cursor = 0
+        self.context['format'] = self._get_format(prefix)
+        if self.errors:
+            return []
+        if not queryset.exists():
+            self.errors.append(oai_errors.NoResults())
+            return []
+        # TODO is there a way to prefetch Sources/Relations/Identifiers just for this slice? https://code.djangoproject.com/ticket/26780
+        works = queryset[cursor:cursor + self.PAGE_SIZE + 1]
```
chrisseto: There is! We just added it into the OSF to fix performance issues. I think I'm going to release it as a pypi package, it seems fairly useful. I'll keep you posted?

chrisseto:
XML templates make a bit uncomfortable. Did you consider looking at something like [this](http://www.django-rest-framework.org/api-guide/renderers/#xml)?

aaxelb:
I looked at DRF renderers but I guess stopped reading halfway through this warning, and templates seemed easy.
> Note: XML output provided is an ad-hoc format that isn't formally described. If you have specific XML requirements you'll need to write your own XML parsers/renderers in order to fully control the representation.

share/oaipmh/views.py
aaxelb: Excellent, I was surprised django didn't have something like that built in.

aaxelb:
I'll write custom renderers, though, it does seem the better way to go.

chrisseto:
❤️  sorry to make you do the extra work.

share/oaipmh/util.py
```diff
@@ -0,0 +1,3 @@
+
+def format_datetime(dt):
+    return dt.strftime('%Y-%m-%dT%H:%M:%SZ')
```
chrisseto: Any reason to not use isoformat?

share/oaipmh/repository.py
```diff
@@ -0,0 +1,175 @@
...
+            metadataRenderer = self._get_metadata_renderer(kwargs['metadataPrefix'])
+            cursor = 0
+        if self.errors:
+            return [], None, None
+        if not queryset.exists():
+            self.errors.append(oai_errors.NoResults())
+            return [], None, None
+        # TODO is there a way to prefetch Sources/Relations/Identifiers just for this slice? https://code.djangoproject.com/ticket/26780
+        works = queryset[cursor:cursor + self.PAGE_SIZE + 1]
+        if len(works) > self.PAGE_SIZE:
```
chrisseto: I might be missing something, won't this always be true until the last page?

```python
In [4]: cursor = 5

In [5]: page_size = 10

In [6]: len(range(100)[cursor:cursor + page_size + 1]) > page_size
Out[6]: True
```

share/oaipmh/util.py
aaxelb: OAI is [picky](https://www.openarchives.org/OAI/openarchivesprotocol.html#Dates) about their dates:
> the "Complete date plus hours, minutes, and seconds" variant of ISO8601. This format is YYYY-MM-DDThh:mm:ssZ.

> When time is included, the special UTC designator ("Z") must be used.

Explicitly formatting it seemed less hacky than isoformat, chop off the end, and add a Z.

share/oaipmh/repository.py
aaxelb: Yep yep, that's how it decides whether it should include a resumption token.

aaxelb:
@chrisseto ListRecords takes about 2.5 seconds per page (100 records), with 15k works in my database. I'm guessing most of that time is due to the lack of prefetching related objects, so it might not get too much worse with more works.

share/oaipmh/repository.py
chrisseto: Ah, makes sense. I think you should be using `.count()` here over `len`. `len` actually loads everything. 

> len(). A QuerySet is evaluated when you call len() on it. This, as you might expect, returns the length of the result list.

aaxelb: Hmm, good point, didn't realize that. `count()` would be an additional query, though, right? Since the queryset is definitely going to be evaluated at this point, it might be best to evaluate it explicitly with `list()` and check the `len` of the result.

PR #610, opened 2017-03-29
[Fix] Only display sources with an icon

leb2dg:


api/views/workflow.py
```diff
@@ -35,7 +35,7 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = SourceSerializer
 
     def get_queryset(self):
-        queryset = Source.objects.all()
+        queryset = Source.objects.exclude(icon='')
```
chrisseto: Should this be `icon__isnull=True`?

share/graphql/query.py
```diff
@@ -41,7 +41,7 @@ def resolve_me(self, args, context, info):
     @graphene.resolve_only_args
     def resolve_sources(self, limit=25, offset=0):
         # TODO Long title will need an index
-        return models.ShareUser.objects.exclude(long_title='').exclude(favicon='').order_by('long_title')[offset:offset + limit]
+        return models.ShareUser.objects.exclude(long_title='').exclude(icon='').order_by('long_title')[offset:offset + limit]
```
chrisseto: See above comment about icon.

Does this result in an `AND` or an `OR` query/ What is it intended to do?

bots/elasticsearch/tasks.py
```diff
@@ -117,8 +117,13 @@ def do_run(self, es_url=None, es_index=None):
     def bulk_stream(self, es_index):
         Source = apps.get_model('share.Source')
         opts = {'_index': es_index, '_type': 'sources'}
-        for source in Source.objects.exclude(long_title='').all():
-            yield {'_op_type': 'index', '_id': source.name, **self.serialize(source), **opts}
+
+        for source in Source.objects.all():
+            # remove sources from search that don't appear on the sources page
+            if source.icon == '':
```
chrisseto: `if source.icon_id is None:` would be preferable here.

`source.icon` will load the entire favicon from the database every time.

share/graphql/query.py
leb2dg: It's an `OR` query (pretty sure). It applies one exclude then the other https://docs.djangoproject.com/en/dev/topics/db/queries/#chaining-filters. Would it be better to use a Q object https://docs.djangoproject.com/en/dev/topics/db/queries/#complex-lookups-with-q-objects?

chrisseto: Looks like `long_title` doesn't exist any more and chaining seems to result in an `AND`

```python
In [6]: str(Source.objects.exclude(long_title='').exclude(icon='').query)
Out[6]: 'SELECT "share_source"."id", "share_source"."name", "share_source"."long_title", "share_source"."home_page", "share_source"."icon", "share_source"."user_id" FROM "share_source" WHERE (NOT ("share_source"."long_title" = ) AND NOT ("share_source"."icon" =  AND "share_source"."icon" IS NOT NULL))'
```

api/views/workflow.py
leb2dg: When clearing an image file through the admin interface, the field isn't nulled. I can't find any documentation on this though?

share/graphql/query.py
aaxelb: `NOT (a OR b) <-> NOT a AND NOT b`

aaxelb: And this should be showing Sources, not ShareUsers

api/views/workflow.py
aaxelb: Might be better to make it just `blank=True`, not `null=True`, and keep this exclude the same

bots/elasticsearch/tasks.py
aaxelb: No, just `if source.icon:` will do. The field returns a little `FieldFile` instance, not the actual favicon.

share/graphql/query.py
chrisseto:
```python
In [2]: str(Source.objects.filter(long_title='foo').exclude(icon='').query)
Out[2]: 'SELECT "share_source"."id", "share_source"."name", "share_source"."long_title", "share_source"."home_page", "share_source"."icon", "share_source"."user_id" FROM "share_source" WHERE ("share_source"."long_title" = foo AND NOT ("share_source"."icon" =  AND "share_source"."icon" IS NOT NULL))'
```

```python
In [3]: str(Source.objects.filter(long_title='foo').filter(icon='1').query)
Out[3]: 'SELECT "share_source"."id", "share_source"."name", "share_source"."long_title", "share_source"."home_page", "share_source"."icon", "share_source"."user_id" FROM "share_source" WHERE ("share_source"."long_title" = foo AND "share_source"."icon" = 1)'
```

bots/elasticsearch/tasks.py
chrisseto: Funky, I stand corrected. 👍

share/graphql/query.py
leb2dg: Ah sorry negative query confusion. I meant ~(P OR Q) which is the same as ~P AND ~Q. Did not know how to see the query though, that's neat.

I didn't look into that method though...seems like the whole thing needs to change if it's being used?

leb2dg: Ah Abram beat me

share/graphql/query.py
```diff
@@ -41,7 +41,7 @@ def resolve_me(self, args, context, info):
     @graphene.resolve_only_args
     def resolve_sources(self, limit=25, offset=0):
         # TODO Long title will need an index
```
aaxelb: While you're in here, this TODO can go away, long_title is uniquely indexed now

api/views/workflow.py
chrisseto: It looks like django will turn this into the correct query? But not with a `.filter`, odd.

```python
In [9]: str(Source.objects.filter(icon='').query)
Out[9]: 'SELECT "share_source"."id", "share_source"."name", "share_source"."long_title", "share_source"."home_page", "share_source"."icon", "share_source"."user_id" FROM "share_source" WHERE "share_source"."icon" = '

In [10]: str(Source.objects.exclude(icon='').query)
Out[10]: 'SELECT "share_source"."id", "share_source"."name", "share_source"."long_title", "share_source"."home_page", "share_source"."icon", "share_source"."user_id" FROM "share_source" WHERE NOT ("share_source"."icon" =  AND "share_source"."icon" IS NOT NULL)'
```

Actually looked at the DB, thought that icon was going to be FK not a string.

@aaxelb's suggest is most correct, I think.

aaxelb: File/ImageFields are weird, and I feel like I abused them pretty hard to store icons in the DB. The string would usually be a path on a filesystem...

chrisseto: You can blame me for that one. 
#TODO

chrisseto:
Will merge once tests pass.

leb2dg:
A 404 is returned when trying to look at a record pushed by a trusted user when that user doesn't have an icon. Looking into it a bit more.

leb2dg:
@chrisseto fixed the 404 thing. Now, if no icon, it looks like:
screen shot 2017-03-29 at 4 12 54 pm

chrisseto:
Wunderbar! waiting on tests

PR #612, opened 2017-03-30
[SHARE-690][Fix] Fix bugs in GWU Preprints Harvester

leb2dg:
*Can switch to hotfix if preferred*
- Remove commas before casting to int
- Update url for request


share/harvesters/edu_gwu.py
```diff
@@ -43,9 +43,8 @@ def fetch_records(self, url, start_date, end_date):
                 break
 
             logger.info('On document %d of %d (%d%%)', count, total, (count / total) * 100)
-
             for link in links:
-                item_response = self.requests.get(self.config.home_page + link)
```
chrisseto: Any reason to not make this the base_url of the config?

leb2dg: Nope, that would be a better solution

PR #614, opened 2017-03-30
Default ingest=True for HarvestLogs

leb2dg:


leb2dg:
@chrisseto, if this is preferred

PR #617, opened 2017-03-31
[Feature] Add --overwrite flag to loadsources

aaxelb:
Change `loadsources`' default behavior to only add sources and source configs that don't exist in the database (based on name/label). With `--overwrite`, it will update all sources/configs to match the yaml.


chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/commit/231f2cb3d4c81921408f80cf00fb8d84b476141c

PR #623, opened 2017-04-10
[SHARE-737][Improvement] Use try advisory lock for harvest tasks

leb2dg:


share/models/ingest.py
```diff
@@ -122,18 +121,11 @@ def get_transformer(self):
...
-                    FROM "{}"
-                    WHERE id = %s
-                    FOR NO KEY UPDATE NOWAIT;
-                '''.format(self._meta.db_table), (self.id,))
-        except DatabaseError:
-            logger.warning('Lock failed; another task is already harvesting %r.', self)
-            raise HarvesterConcurrencyError('Unable to lock {!r}'.format(self))
-        else:
+        with connections[using].cursor() as cursor:
+            cursor.execute('''SELECT pg_try_advisory_lock('{}'::regclass::integer, {});'''.format(self._meta.db_table, self.id))
```
chrisseto: `self.id` should be passed as a parameter to `cursor.execute` to fill in a `%s`. Actually, it looks like you could do that with `db_table` as well.

Any reason that you're double casting (`::regclass::integer`) `self.id`?

chrisseto: Do you know if this will perform a [session level lock or a transaction level lock](https://www.postgresql.org/docs/9.4/static/explicit-locking.html#ADVISORY-LOCKS)?

leb2dg: Oh we want transaction level don't we? `pg_try_advisory_lock` performs a session level lock but [pg_try_advisory_xact_lock](https://www.postgresql.org/docs/9.5/static/functions-admin.html#FUNCTIONS-ADVISORY-LOCKS) performs a transaction level lock.

leb2dg: Is there double casting? The first cast is for the table name and `self.id` is not cast.

PR #624, opened 2017-04-11
[Hotfix] 2.6.8

chrisseto:


api/views/workflow.py
```diff
@@ -35,8 +36,8 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     serializer_class = SourceSerializer
 
     def get_queryset(self):
-        queryset = Source.objects.exclude(icon='')
-        sort = self.request.query_params.get("sort")
+        queryset = Source.objects.exclude(models.Q(icon='') | models.Q(is_active=False))
```
chrisseto: I'll add tests for graphql.
Sometimes I hate boolean logic, 👍  will change.

chrisseto:
@leb2dg is_active will stop them from showing up on the sources page and the sources ES index. Should I go ahead and exclude them from the sources field on creative works?

leb2dg:
As a user, if there is data from a source, I would expect the source to be listed/shown. If there is no data for a source, I would not expect the source to be shown/listed.

I guess my question is what use cases is the `is_active` flag covering?

chrisseto:
@leb2dg `is_active` is being added to avoid having to delete Biorxiv RSS. Does that answer your question?

share/management/commands/loadsources.py
```diff
@@ -23,8 +21,14 @@ def add_arguments(self, parser):
         parser.add_argument('--overwrite', action='store_true', help='Overwrite existing sources and source configs')
 
     def handle(self, *args, **options):
+        global apps
```
aaxelb: Eeeeeeew. But fair enough, it has to work somehow...

bots/elasticsearch/tasks.py
```diff
@@ -120,7 +120,7 @@ def bulk_stream(self, es_index):
 
         for source in Source.objects.all():
             # remove sources from search that don't appear on the sources page
-            if not source.icon:
+            if not source.icon or not source.is_active:
```
aaxelb: `is_deleted`

share/management/commands/loadsources.py
chrisseto: It's probably a better idea to just tack this onto `self`. Not sure why I didn't think of that first.

aaxelb: Yeah, that is better.

PR #627, opened 2017-04-17
[SHARE-760][Feature] Allow easy creation of Sources from the admin

chrisseto:


leb2dg:
I'm getting a `ValueError` when trying to save duplicate Sources. `'SourceForm' has no field named 'long_title'.`

share/admin.py
```diff
@@ -353,8 +355,58 @@ class ShareUserAdmin(admin.ModelAdmin):
...
+    def validate_unique(self):
+        # Forces validation checks on every field
+        try:
+            self.instance.validate_unique()
+        except forms.ValidationError as e:
+            self._update_errors(e)
+
+    def _post_clean(self):
+        self.cleaned_data['long_title'] = self.cleaned_data.pop('name')[0]
+        self.cleaned_data['name'] = '.'.join(reversed(self.cleaned_data['home_page'].split('//')[1].split('.')))
```
leb2dg: This will allow for `www` at the end of source `long_title`'s which isn't a blocker just slightly annoying.

share/admin.py
```diff
@@ -353,8 +355,58 @@ class ShareUserAdmin(admin.ModelAdmin):
...
+
+    class Meta:
+        model = Source
+        exclude = ('user', 'long_title', )
+        widgets = {'name': TextInput}
+
+    def validate_unique(self):
+        # Forces validation checks on every field
+        try:
+            self.instance.validate_unique()
```
leb2dg: This breaks the page if `long_title` isn't unique

share/admin.py
```diff
@@ -353,8 +355,58 @@ class ShareUserAdmin(admin.ModelAdmin):
...
+
+    def _post_clean(self):
+        self.cleaned_data['long_title'] = self.cleaned_data.pop('name')[0]
+        self.cleaned_data['name'] = '.'.join(reversed(self.cleaned_data['home_page'].split('//')[1].split('.')))
+        return super()._post_clean()
+
+    def save(self, commit=True):
+        instance = super().save(commit=False)
+
+        user = ShareUser.objects.create_user(username='sources.' + instance.name, save=commit)
```
leb2dg: It's a little confusing that this also adds a user. May just want to make it more clear on the source form?

share/admin.py
chrisseto: It's supposed to, an error message will be sent back to whomever submitted the form.

leb2dg: There isn't a field on the form for `long_title` though so it just blows up. Is that intended?

share/admin.py
chrisseto: What would make it more clear? A user is required for a source to be made and its a bit nicer if names match up.

share/admin.py
chrisseto: Oh, sorry I didn't see your above comment. I only handled conflicting names. 👍  Nice catch.

share/admin.py
leb2dg: Maybe just a note on the form. If we are the only people using it then it doesn't really need to be more clear but if product or others were to create a source then the implicit user creation would not be obvious.

chrisseto:
@leb2dg added a couple notes to the form and fixed the value error.

For some reason access_token shows up on the add page but is uneditable. I don't think it's worth fighting with django to remove it.

leb2dg:
@chrisseto, might want some more error handling:
- KeyError
    - no url/title is given
    - url with no `.` or `\\`
- Error highlights title field when the url is a duplicate (less important)

Otherwise looks good and works as expected 🥇

chrisseto:
@leb2dg There's a test case for everything I could think of 👍

leb2dg:
It's beautiful 🌈

PR #629, opened 2017-04-19
[SHARE-746][Improvement] Change ports for rabbitmq, postgres, and elasticsearch

leb2dg:
## Purpose
Make it easier to have both SHARE and the OSF running locally.

## Changes
Added 30,000 to all SHARE services ports:
* postgres: 5432 --> 35432
* elasticsearch: 9200 --> 39200
* rabbitmq: 5672 --> 35672
* web: 4200 -->34200

Related PR: https://github.com/CenterForOpenScience/ember-share/pull/191

This could be done in other ways:
* Adding SHARE to OSF docker-compose
    * more likely to be forgotten about/get out of date
* Using the same containers as OSF
    * I don't think this would work very well from what little I attempted
    * would depend on us using the same images as the OSF

## Side effects
Will require some changes to staging/production. Also changes to quay images.

Could just have a separate docker-compose.yml for local though?

## Ticket
https://openscience.atlassian.net/browse/SHARE-746


docker-compose.yml
```diff
@@ -60,15 +60,15 @@ services:
...
-      SHARE_API_URL: http://web:8000/
-      ELASTICSEARCH_URL: http://elasticsearch:9200/
+      BROKER_URL: amqp://guest:guest@rabbitmq:35672/
+      SHARE_API_URL: http://web:38000/
+      ELASTICSEARCH_URL: http://elasticsearch:39200/
 
   web:
     build: .
-    command: python manage.py runserver --noreload 0.0.0.0:8000
+    command: python manage.py runserver --noreload 0.0.0.0:38000
```
chrisseto: Does this work? You've told the server in the container to run on port 38000 but you're forwarding the host's port 8000 in the container still.

https://docs.docker.com/compose/compose-file/#ports

docker-compose.yml
```diff
@@ -60,15 +60,15 @@ services:
       - ./:/code
     environment:
       DATABASE_HOST: postgres
-      BROKER_URL: amqp://guest:guest@rabbitmq:5672/
-      SHARE_API_URL: http://web:8000/
-      ELASTICSEARCH_URL: http://elasticsearch:9200/
+      BROKER_URL: amqp://guest:guest@rabbitmq:35672/
```
chrisseto: Wouldn't this need to be kept as the default port? The actual connection should be happening inside the container.

manage.py
```diff
@@ -12,6 +14,11 @@
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "project.settings")
 
+    # Override default port for `runserver` command
+    django.setup()
+    from django.core.management.commands.runserver import Command as runserver
+    runserver.default_port = "38000"
```
chrisseto: This feels very hacky. Could you overwrite the runserver command with your own and then just subclass it?

laurenbarker:
@aaxelb, conflicts resolved. The docs/`README.d`/`sharectl server` need to be updated though.

PR #630, opened 2017-04-19
[SHARE-563][Bug] Handle scroll api 500s

laurenbarker:
## Purpose
Handle elasticsearch scroll API requests.

## Changes
~~* Return `501` response if scroll requested
    * *also considered 400 and 403, thoughts?*~~
* Return `403` response if any improper search request is made 
* Add tests for es scroll API

## Side effects
None

## Ticket
https://openscience.atlassian.net/browse/SHARE-563

### DIFF
--- a/api/urls.py
+++ b/api/urls.py
@@ -98,7 +98,14 @@ urlpatterns = [
     url(r'atom/?', views.CreativeWorksAtom(), name='atom'),
     url(r'graph/?', GraphQLView.as_view(graphiql=True)),
     url(r'userinfo/?', ensure_csrf_cookie(views.ShareUserView.as_view()), name='userinfo'),
-    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
+
+    # only match _count, _search, _suggest requests
+    url(
+        r'search/(?!.*_bulk\/?$)(?P<url_bits>(_|[a-zA-Z]+/_)((?=search)|(?=count)|(?=suggest))[a-zA-Z]{5,7}/{0,1}$)',
+        csrf_exempt(views.ElasticSearchView.as_view()),
+        name='search'
+    ),
+    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearch403View.as_view()), name='search_403'),
 
     url(r'schema/?$', views.SchemaView.as_view(), name='schema'),
     url(r'schema/creativework/hierarchy/?$', views.ModelTypesView.as_view(), name='modeltypes'),
--- a/api/urls.py
+++ b/api/urls.py
@@ -99,15 +99,17 @@ urlpatterns = [
     url(r'graph/?', GraphQLView.as_view(graphiql=True)),
     url(r'userinfo/?', ensure_csrf_cookie(views.ShareUserView.as_view()), name='userinfo'),
 
-    # only match _count, _search, _suggest requests
+    # only match _count and _search requests
     url(
-        r'search/(?!.*_bulk\/?$)(?P<url_bits>(_|[a-zA-Z]+/_)((?=search)|(?=count)|(?=suggest))[a-zA-Z]{5,7}/{0,1}$)',
+        r'search/(?P<url_bits>(_|[a-zA-Z]+/_)(?:search|count)/{0,1}$)',
         csrf_exempt(views.ElasticSearchView.as_view()),
         name='search'
     ),
+    # match _suggest requests
+    url(r'search/(?P<url_bits>(_|[a-zA-Z]+/_)(?:suggest)/{0,1}$)', csrf_exempt(views.ElasticSearchPostOnlyView.as_view()), name='search_post'),
     # match _mappings requests
-    url(r'search/(?P<url_bits>_mappings(/.+|$|/))', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
-    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearch403View.as_view()), name='search_403'),
+    url(r'search/(?P<url_bits>_mappings(/.+|$|/))', csrf_exempt(views.ElasticSearchGetOnlyView.as_view()), name='search_get'),
+    url(r'search/(?P<url_bits>.*)', csrf_exempt(views.ElasticSearch403View.as_view()), name='search_403'),
 
     url(r'schema/?$', views.SchemaView.as_view(), name='schema'),
     url(r'schema/creativework/hierarchy/?$', views.ModelTypesView.as_view(), name='modeltypes'),
--- a/api/urls.py
+++ b/api/urls.py
@@ -101,15 +101,27 @@ urlpatterns = [
 
     # only match _count and _search requests
     url(
-        r'search/(?P<url_bits>(_|[a-zA-Z]+/_)(?:search|count)/{0,1}$)',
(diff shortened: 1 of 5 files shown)


chrisseto:
I would go with either a 400 or 403.
I think a 403 would be more correct, if you go with that would you mind changing the existing 400s to match?

chrisseto:
It might be a touch easier to add a URL pattern with high priority than the other ES url that always sends back a 403. It would be a bit more explicit and, I think, easier to maintain.

leb2dg:
@chrisseto, does it make more sense to only match valid requests and then catch the rest? There are less valid cases. Or would this implementation cause problems/are there valid cases missing from the tests?

chrisseto:
@leb2dg That's a much better idea. I think most of the valid cases have tests? If not, the frontend and OSF's count in search are the only endpoints we need to worry about.

leb2dg:
@chrisseto, double checked the frontend and added `_mappings`.

tests/api/test_elasticsearch.py
```diff
@@ -26,18 +26,24 @@ def test_cannot_access_bulk(self, client):
...
-        assert client.post('/api/v2/search/type/id/').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else/').status_code == 400
-        assert client.post('/api/v2/search/type/id/_search').status_code == 400
-        assert client.post('/api/v2/search/type/id/_search/').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else/_search').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else/_search/').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else/_count').status_code == 400
-        assert client.post('/api/v2/search/type/id/some/thing/else/_count/').status_code == 400
+        assert client.post('/api/v2/search/type').status_code == 403
```
chrisseto: Could you add tests for `_mappings` and maybe some random garbage urls?

The garbage URLs would be psuedo fuzz testing as well as validating that we are whitelisting not blacklisting.

api/urls.py
```diff
@@ -98,7 +98,16 @@ def register_url(self, subclass, viewset):
...
-    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
+
+    # only match _count, _search, _suggest requests
+    url(
+        r'search/(?!.*_bulk\/?$)(?P<url_bits>(_|[a-zA-Z]+/_)((?=search)|(?=count)|(?=suggest))[a-zA-Z]{5,7}/{0,1}$)',
+        csrf_exempt(views.ElasticSearchView.as_view()),
+        name='search'
+    ),
+    # match _mappings requests
+    url(r'search/(?P<url_bits>_mappings(/.+|$|/))', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
```
chrisseto: Any reason this isn't included in the above route?

api/urls.py
```diff
@@ -98,7 +98,16 @@ def register_url(self, subclass, viewset):
     url(r'atom/?', views.CreativeWorksAtom(), name='atom'),
     url(r'graph/?', GraphQLView.as_view(graphiql=True)),
     url(r'userinfo/?', ensure_csrf_cookie(views.ShareUserView.as_view()), name='userinfo'),
-    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
+
+    # only match _count, _search, _suggest requests
+    url(
+        r'search/(?!.*_bulk\/?$)(?P<url_bits>(_|[a-zA-Z]+/_)((?=search)|(?=count)|(?=suggest))[a-zA-Z]{5,7}/{0,1}$)',
```
chrisseto: Could you simplify this regex a bit?

if you don't need to I would recommend against positive look aheads, they're [a bit odd](http://www.regular-expressions.info/lookaround.html).
This would be much more clear as `(?:_search|_count|_suggest)`

I think you can remove the `_bulk` now that this is a whitelist endpoint.

If it is easier to understand and/or other easier to handle, it might make sense to separate out endpoints that we allow POSTing, GETTing, or POST and GET?
Just an off hand idea.

api/urls.py
laurenbarker: `_mappings` requests have a different format then the other whitelisted urls. I can try and put it in the same regex but I think it would be messy.

api/urls.py
laurenbarker: Separating them out would probably be the best thing to do 👍

tests/api/test_elasticsearch.py
laurenbarker: @chrisseto, I'm not really sure what you had in mind for "pseudo fuzz testing". Would it be similar to the existing tests or more like a random url generator?

api/urls.py
```diff
@@ -98,7 +98,18 @@ def register_url(self, subclass, viewset):
     url(r'atom/?', views.CreativeWorksAtom(), name='atom'),
     url(r'graph/?', GraphQLView.as_view(graphiql=True)),
     url(r'userinfo/?', ensure_csrf_cookie(views.ShareUserView.as_view()), name='userinfo'),
-    url(r'search/(?!.*_bulk\/?$)(?P<url_bits>.*)', csrf_exempt(views.ElasticSearchView.as_view()), name='search'),
+
+    # only match _count and _search requests
+    url(
+        r'search/(?P<url_bits>(_|[a-zA-Z]+/_)(?:search|count)/{0,1}$)',
```
chrisseto: `{0,1}` is equivalent to `?`, 0 or one of the preceding element.

chrisseto: `[a-zA-Z]` can be `\w`. That would match `0-9` and `_` as well, but I think that's okay, if not desirable.

chrisseto: I think this might actually break? `search` and `count` are outside of the `url_bits` capture group. The view will only end up seeing things preceding `search` or `count`.

You'll probably want something closer to this?

`(?P<url_bits>(?:\w+/)?_(?:search|count)/?)$`

* `(?P<url_bits>` - Start a named capture group
* `(?:\w/+)?` - Optionally match one or more alpha-numeric characters followed by a `/`, without capturing that group
* `_(?:search|count)` - Match either `_search` or `_count`
* `/?` - Optionally match a `/`
* `)` - Close the named capture group
* `$` - Matches the end of the input. Required the last character to be either `/`, the `h` from `search`, or the `t` from `count`

chrisseto:
This might also be a good time to add a test that hitting this endpoint actually proxies correctly to ES. There's a fixture that will provision ES for you, if it is available.

api/urls.py
laurenbarker: `search` and `count` are in the capture group `( ?P<url_bits> ( _|[a-zA-Z]+/_ ) ( ?:search|count ) /{0,1}$ )` but your regex is definitely better

chrisseto: Oh dang, you're right. Sorry about that!

tests/api/test_elasticsearch.py
chrisseto: Either way would be fine, hard coded or generated. Just some URLs with various amounts of character, `/`s, or prepended with `_`s.

Things like:
`/api/v2/search/__count`
`/api/v2/search/_counttttttttttt`
`/api/v2/search/_coun`

chrisseto:
Adendum to the above though, you could also just make asserts about what requests.get/post gets called with.

maxogden:
Why is the https://share.osf.io/api/v2/search/_search/?scroll=1m API being deprecated here?

PR #631, opened 2017-04-21
[SHARE-761][Feature] Add sources that have not submitted new changes

chrisseto:


share/tasks.py
```diff
@@ -383,21 +384,21 @@ def do_run(self, *args, **kwargs):
...
-        logger.info('%s started make JSON patches for NormalizedData %s at %s', self.started_by, self.normalized.id, datetime.datetime.utcnow().isoformat())
+        logger.info('%s started disambiguation for NormalizedData %s at %s', self.started_by, self.normalized.id, datetime.datetime.utcnow().isoformat())
 
         try:
             with transaction.atomic():
                 cg = ChangeGraph(self.normalized.data['@graph'], namespace=self.normalized.source.username)
                 cg.process()
                 cs = ChangeSet.objects.from_graph(cg, self.normalized.id)
-                if cs and (self.source.is_robot or self.source.is_trusted):
+                if cs and (self.source.is_robot or self.source.is_trusted or Source.objects.filter(user=self.source).exists()):
```
laurenbarker: Does this make it so if there is normalized data from a source in the database any new change sets will be accepted automatically?

laurenbarker: Oh wait, is this because a Source object will only exist for trusted users or robots? Do they all need to be checked?

chrisseto:
This was hotfixed into develop

PR #632, opened 2017-04-21
[SHARE-743][Feature] Django 1.11

aaxelb:
- Upgrade Django to 1.11
- Upgrade Django-related dependencies
- Fix anything that broke
- Resolve all deprecation warnings in our code


chrisseto:
@aaxelb Looks great so far. Could you replace all the newly added `DATABASE_CASCADES` with Django's cascade? We won't have to do any migrations in that case 👍

share/models/creative.py
```diff
@@ -39,7 +39,7 @@ def normalize(self, node, graph):
     class Disambiguation:
         any = ('identifiers',)
 
-    class Meta:
+    class Meta(ShareObject.Meta):
```
chrisseto: What is this doing? I haven't seen anything in the release notes that would warrant doing this.

Does the `abstract = True` mess things up at all?

share/models/base.py
```diff
@@ -166,11 +144,11 @@ class Meta:
 class ShareObject(models.Model, metaclass=ShareObjectMeta):
     id = models.AutoField(primary_key=True)
     objects = ShareObjectManager()
-    versions = VersionManager()
     changes = fields.GenericRelationNoCascade('Change', related_query_name='share_objects', content_type_field='target_type', object_id_field='target_id', for_concrete_model=True)
 
     class Meta:
         abstract = True
+        base_manager_name = 'objects'
```
chrisseto: Any reason to do this as rather than setting `_base_manager`?

Is this why all the other Metas inherit from this one?

share/migrations/0030_auto_20170425_1427.py
```diff
@@ -0,0 +1,194 @@
+# -*- coding: utf-8 -*-
```
chrisseto: Regenerate this please.

share/models/base.py
aaxelb: `base_manager_name` is the [recommended](https://docs.djangoproject.com/en/1.11/topics/db/managers/#using-managers-for-related-object-access) way to have a custom `_base_manager`. Setting `_base_manager` directly might work too, but that seems like a rug that could be pulled out from under us in a future upgrade.

And yes, that's why the Metas inherit.

share/models/creative.py
aaxelb: It seemed cleaner to have them inherit than set `base_manager_name` all over.

`Meta.abstract` is a [special case](https://docs.djangoproject.com/en/1.11/topics/db/models/#meta-inheritance) and is not inherited.

chrisseto: 👍  Thank you for explain.

share/migrations/0030_auto_20170425_1427.py
aaxelb: I did, the only changes are Meta options and setting `related_name='versions'` on all the `persistent_ids`. 1.11 (or maybe 1.10, I forget) was unhappy with the `versions` manager on model instances, and when I was halfway through trying to rehack it I realized there was no reason not to just use the related manager instead.

aaxelb: (though do correct me if I missed something)

chrisseto: Ah, got it.

chrisseto: Nope, I'm just bad at reading things in their entirety

aaxelb:
@chrisseto We don't have any custom migration operations (just various RunPython operations invoking our code), so I don't think clear_delayed_apps_cache affects us.

PR #638, opened 2017-04-25
[SHARE-820][Task] Disable IUScholarWorks harvester

laurenbarker:
### Purpose
Disable IUScholarWorks harvester because it is returning 500s.

Disabled manually on production.


laurenbarker:
They fixed their endpoint. Enabled on production.

PR #640, opened 2017-04-27
Hotfix/2.6.12

chrisseto:


laurenbarker:
Looks good to me; functions as expected.

Just to confirm I understand. You are subclassing the djcelery DatabaseBackend and calling the same functions but catching all exceptions. If multiple exceptions are thrown raise `SystemExit` so that cleanup can still occur. 

[Reference](https://docs.python.org/3/library/sys.html):
> This is implemented by raising the SystemExit exception, so cleanup actions specified by finally clauses of try statements are honored, and it is possible to intercept the exit attempt at an outer level.

The travis errors are from the same elastic thing that will be fixed in 2.7 release?

chrisseto:
> Just to confirm I understand: You are subclassing the djcelery DatabaseBackend and calling the same functions but catching all exceptions. 

Spot on.

> If multiple exceptions are thrown raise SystemExit so that cleanup can still occur.

Not quite for this one. The extra try there is just me being paranoid. If any exception is raised, kill the process via the `SystemExit` exception.

> The travis errors are from the same elastic thing that will be fixed in 2.7 release.

Look to be 👍

PR #641, opened 2017-04-28
Hotfix/2.6.13

chrisseto:


share/change.py
```diff
@@ -292,18 +292,30 @@ def change(self):
...
+                    # Just in case
+                    ignore_attrs.update(self.instance.change.change.keys())
+
+                    # Go back until we find a change that is older than us
+                    for version in self.instance.versions.select_related('change').all():
+                        if not version.date_updated or date_updated > version.date_updated:
+                            break
+                        ignore_attrs.update(version.change.change.keys())
+
+            # Pretty self-explanatory. If we get changes from a source that hasn't been marked as canonical
```
laurenbarker: > Pretty self-explanatory.

is just extra text and could be removed.

share/change.py
```diff
@@ -292,18 +292,30 @@ def change(self):
...
+            # Don't allow attributes set by canonical sources to be changed.
+            # Stops aggregators from providng more generic sources or overwriting changes that have been pushed to use.
```
laurenbarker: > Stops aggregators from providng more generic sources or overwriting changes that have been pushed to use.

This sentence isn't clear. I'm not sure what "providing more generic sources" means or what "changes that have been pushed to use" means. Would "Stops aggregators from overwriting attributes set by canonical sources." cover the meaning? If so, we might not need the last sentence.

share/change.py
```diff
@@ -292,18 +292,30 @@ def change(self):
         # Hacky fix for SHARE-604
         # If the given date_updated is older than the current one, don't accept any changes that would overwrite newer changes
         ignore_attrs = set()
-        if issubclass(self.model, apps.get_model('share', 'creativework')) and 'date_updated' in self.attrs and self.instance.date_updated:
-            date_updated = pendulum.parse(self.attrs['date_updated'])
```
laurenbarker: Has this broken before?

share/management/commands/loadsources.py
```diff
@@ -132,7 +133,11 @@ def schedule_harvest_task(self, label, disabled):
     def process_defaults(self, model, defaults):
         ret = {}
         for k, v in defaults.items():
-            field = model._meta.get_field(k)
+            try:
```
laurenbarker: This is to handle the `canonical` field set in the source config that doesn't actually do anything?

tests/share/disambiguation/test_creativework.py
```diff
@@ -259,3 +260,96 @@ def test_no_timetraveling_many(self, Graph):
...
+        other_graph.process()
+        ChangeSet.objects.from_graph(other_graph, NormalizedDataFactory(source=nc_source.user).id).accept()
+
+        cw.refresh_from_db()
+
+        assert other_graph.nodes[0].change == {}
+        assert cw.type == 'share.preprint'
+        assert cw.title == 'The first title'
+
+    def test_canonical_overwrites_canonical(self, Graph):
```
laurenbarker: > If more than one canonical source contribute to a single document all incoming change will fail.

This comment in `ingest.py` is inconsistent with the functionality.

tests/share/disambiguation/test_creativework.py
```diff
@@ -259,3 +260,96 @@ def test_no_timetraveling_many(self, Graph):
 
         older_graph.process()
         assert older_graph.nodes[0].change == {'description': 'The final description'}
+
+    def test_cannot_overwrite_cannonical(self, Graph):
```
laurenbarker: Could you add a test for canonical overwrites not canonical so `canonical=False` then `canonical=True`. That way all cases are covered.

tests/share/models/test_change.py
```diff
@@ -104,115 +104,6 @@ def test_update_creative_work(self, change_factory):
         assert change_set.changes.count() == 1
         assert change_set.changes.first().target == preprint
 
-    # def test_update_requires_saved(self, share_source):
```
laurenbarker: Woo!

laurenbarker:
There are also a couple of tests that are failing that weren't before:
`tests/share/normalize/test_normalize.py:250: `
`tests/api/test_generated_endpoints.py:136: `

share/change.py
chrisseto: Not that I know of? The extra check for `self.instance` can be removed. I forgot about the early return on line 274.

share/management/commands/loadsources.py
chrisseto: This script gets run by a migration before the canonical field is created. This just ignores fields that don't exist when this script is run. I'll add a note.

tests/share/disambiguation/test_creativework.py
```diff
@@ -259,3 +260,44 @@ def test_no_timetraveling_many(self, Graph):
...
+        assert cw.type == 'share.preprint'
+        assert cw.title == 'The first title'
+
+        second_graph.process()
+        ChangeSet.objects.from_graph(second_graph, NormalizedDataFactory(source=second_source.user).id).accept()
+
+        cw = models.AbstractCreativeWork.objects.get(id=cw.id)
+
+        assert second_graph.nodes[0].change == change
+        assert cw.type == change.get('type', 'share.preprint')
```
laurenbarker: I feel like it weird to set a default instead of defining each one in the parameters; especially since the default only should occur once.

share/tasks.py
```diff
@@ -383,21 +384,21 @@ def do_run(self, *args, **kwargs):
...
-        logger.info('%s started make JSON patches for NormalizedData %s at %s', self.started_by, self.normalized.id, datetime.datetime.utcnow().isoformat())
+        logger.info('%s started disambiguation for NormalizedData %s at %s', self.started_by, self.normalized.id, datetime.datetime.utcnow().isoformat())
 
         try:
             with transaction.atomic():
                 cg = ChangeGraph(self.normalized.data['@graph'], namespace=self.normalized.source.username)
                 cg.process()
                 cs = ChangeSet.objects.from_graph(cg, self.normalized.id)
-                if cs and (self.source.is_robot or self.source.is_trusted):
+                if cs and (self.source.is_robot or self.source.is_trusted or Source.objects.filter(user=self.source).exists()):
```
laurenbarker: I commented on this in the original PR. Are all 3 checks needed? Isn't `Source.objects.filter(user=self.source).exists()` equivalient to `self.source.is_robot or self.source.is_trusted`?

share/models/ingest.py
```diff
@@ -73,6 +73,14 @@ class Source(models.Model):
     icon = models.ImageField(upload_to=icon_name, storage=SourceIconStorage(), blank=True)
     is_deleted = models.BooleanField(default=False)
 
+    # Whether or not this SourceConfig collects original content
+    # If True changes made by this source cannot be overwritten
```
laurenbarker: I suppose `cannot be overwritten` should now say `cannot be overwritten by a non-canonical source`

share/management/commands/loadsources.py
laurenbarker: 👍

share/change.py
```diff
@@ -276,21 +289,40 @@ def change(self):
...
+            if self.graph.source and not self.graph.source.canonical:
+                prev_changes = list(self.instance.changes.filter(change_set__normalized_data__source__source__canonical=True).values_list('change', flat=True))
+                canonical_keys = set(key for change in prev_changes for key in change.keys())
+                if prev_changes and set(self.attrs.keys()) & canonical_keys:
+                    canonical_sources = list(self.instance.sources.filter(source__canonical=True).values_list('username', flat=True))
+                    logger.warning('Recieved changes from a non-canonical source %s that conflict with one of %s. Ignoring conflicting changes', self.graph.source, canonical_sources)
+                    ignore_attrs.update(canonical_keys)
+
+                    # Appears that type doesn't get added to changes or at least the first change
+                    # Safe to assume the type was set by the canonical source
```
laurenbarker: I'm confused, in the tests the type is reassigned?

laurenbarker:
Still one failing travis test

share/tasks.py
chrisseto: They are all different cases, technically.  I added `Source.objects.filter(user=self.source).exists()` so we could start moving away from the others.

PR #642, opened 2017-05-01
[SHARE-816][Fix] Handle marked as deleted creative work  

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-816

## Problem

Currently, when something is sent to SHARE and marked as deleted it will be removed from elasticsearch. However, this data will still be returned via the API. The API should send back a 403 if something is marked as is_deleted
## Solution:

- Update URL EndpointGenerator to exclude items with is_deleted field equals to true.
- Replace the get_object_or_404 with get_object_or_404_or_403 to check if item is in database or not and based on that returns the object, 404 or 403.


chrisseto:
@sheriefvt for future reference would you mind formatting the title of PRs and commit messages as :

```
[SHARE-XXX][Feature/Fix/Task] Brief description

    * Additional information about this fix
      like side effects.
    * Preferably formatted as a bulleted list.
```

Lines should wrap at around 80 characters.
If a line does wrap line up the first letter with the first letter of the line it is extending.

api/urls.py
```diff
@@ -58,6 +58,9 @@ def generate_viewset(self, subclass, serializer):
         if subclass.__name__ == 'AgentIdentifier':
             queryset = queryset.exclude(scheme='mailto')
 
+        if 'is_deleted' in [field.name for field in serializer.Meta.model._meta.get_fields()]:
```
aaxelb: Instead of iterating through all the fields, I think it would be cleaner and faster to try `get_field('is_deleted')` and catch `FieldDoesNotExist`

api/views/share.py
```diff
@@ -2,18 +2,66 @@
...
+    arguments and keyword arguments are used in the get() query.
+
+    Note: Like with get(), an MultipleObjectsReturned will be raised if more than one
+    object is found.
+    """
+    queryset = _get_queryset(klass)
+    try:
+        return queryset.get(*args, **kwargs)
+    except queryset.model.DoesNotExist:
+        if AbstractCreativeWork.objects.filter(id=kwargs['pk']).count() == 1:
```
aaxelb: `kwargs` won't always contain `'pk'` (like in `source_icon_view` below). Since you can't assume what filters are passed in, it might be better to rework this so you don't have to.

api/views/share.py
```diff
@@ -2,18 +2,66 @@
...
+                         "Manager, or QuerySet" % klass__name)
+    return manager.all()
+
+
+def get_object_or_404_or_403(klass, *args, **kwargs):
+    """
+    Replaces the get_object_or_404 in django.shortcuts.
+
+    Uses get() to return an object, or raises a Http404 exception if the object
+    does not exist.
```
aaxelb: The docstring should also mention when/why `PermissionDenied` is raised.

api/views/share.py
```diff
@@ -2,18 +2,66 @@
...
+        if isinstance(klass, type):
+            klass__name = klass.__name__
+        else:
+            klass__name = klass.__class__.__name__
+        raise ValueError("Object is of type '%s', but must be a Django Model, "
+                         "Manager, or QuerySet" % klass__name)
+    return manager.all()
+
+
+def get_object_or_404_or_403(klass, *args, **kwargs):
```
chrisseto: This would be better implemented in the ViewSet, either in `get_object`, `check_object_permissions`, or both.

Django Rest Framework sticks to OOP style coding. Rather than making helper methods that sit in the aether, look for the appropriate method to override or hook to implement.

api/views/share.py
```diff
@@ -45,6 +46,12 @@ def get_object(self):
         except InvalidID:
             raise serializers.ValidationError('Invalid ID')
 
+        try:
```
chrisseto: Try popping this logic into a [custom permissions class](http://www.django-rest-framework.org/api-guide/permissions/#custom-permissions).

`if hasattr(object, attrname):` is a bit more readable and more safe than catching an `AttributeErrror`.
Anything could raise an `AttributeError`, not just the attribute you're checking for.

api/permissions.py
```diff
@@ -27,3 +29,15 @@ def has_permission(self, request, view):
...
                        '`oauth2_provider.rest_framework.OAuth2Authentication` authentication '
                        'class to be used.')
+
+
+class IsDeletedPremissions(BasePermission):
+    """
+    Permission check for deleted objects.
+    """
+    def has_object_permission(self, request, view, obj):
+        TargetObject = AbstractCreativeWork.objects.filter(id=obj.id).first()
```
chrisseto: There no need to reload the object from the database. You can just check `obj.is_deleted`.

Variables should always be lowercase and snakecased, use `target_object` rather than `TargetObject`.

api/permissions.py
```diff
@@ -27,3 +29,15 @@ def has_permission(self, request, view):
...
+        if hasattr(TargetObject, 'is_deleted'):
```
chrisseto: As there are no other code paths, this would be better expressed using `and`. 
if hasattr(...) and ...is_deleted:`

api/views/share.py
```diff
@@ -53,9 +55,24 @@ def get_object(self):
 
         return obj
 
+    # Override to filter out deleted items
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        try:
+            queryset = queryset.filter(is_deleted=False)
+        except:
```
chrisseto: Bare except blocks are very dangerous. This will catch and swallow any exception, included KeyboardInterrupts.

Catch the `FieldError` that would be raised here.

api/views/share.py
```diff
@@ -53,9 +55,24 @@ def get_object(self):
 
         return obj
 
+    # Override to filter out deleted items
+    def list(self, request, *args, **kwargs):
+        queryset = self.filter_queryset(self.get_queryset())
+        try:
+            queryset = queryset.filter(is_deleted=False)
```
chrisseto: This would be better implemented by overriding `self.filter_queryset`

api/permissions.py
```diff
@@ -1,6 +1,8 @@
 from oauth2_provider.ext.rest_framework import TokenHasScope
```
chrisseto: Imports should be ordered alphabetically and separated into stdlib, 3rdparty, and application.

```python
import abc
import os

import django
import requests

import share.models
```

Looks like this file is a bit messed up already, would you mind reordering it?

api/urls.py
```diff
@@ -1,6 +1,7 @@
 from typedmodels.models import TypedModel
 
 from django.conf.urls import url
+
```
chrisseto: No need for the additional newline here.

api/views/share.py
```diff
@@ -19,13 +22,35 @@
...
+
     # Can't expose these until we have indexes added, both ascending and descending
     # filter_backends = (filters.OrderingFilter,)
     # ordering_fields = ('id', 'date_updated')
 
-    # override to convert encoded pk to an actual pk
+    # Override get_queryset to handle items marked as deleted.
+    def get_queryset(self, list=True):
+
+        assert self.queryset is not None, (
```
chrisseto: Rather than copying the base method, just call `super().get_queryset()`

api/views/share.py
```diff
@@ -19,13 +22,35 @@
...
+                try:
+                    queryset = queryset.filter(is_deleted=False)
+                except FieldError:
+                    pass
+        return queryset
+
+    # Override to convert encoded pk to an actual pk
     def get_object(self):
-        queryset = self.filter_queryset(self.get_queryset())
+        queryset = self.filter_queryset(self.get_queryset(False))
```
chrisseto: Is there any reason to change this?

tests/api/test_generated_endpoints.py
```diff
@@ -94,6 +95,45 @@ def test_get_data(self, generator, model, route, controlled_values, client, Grap
         for value in controlled_values:
             assert actual['data']['attributes'][value] == expected[camelCase_to_underscore(value)]
 
+    def test_can_delete_work(self, client, normalized_data_id):
```
chrisseto: Prefer to use classes to logically group your tests together.

tests/api/test_generated_endpoints.py
```diff
@@ -94,6 +95,45 @@ def test_get_data(self, generator, model, route, controlled_values, client, Grap
...
+        graph = ChangeGraph([{
```
chrisseto: Rather than building up the entire change graph, can you just call `factories.AbstractCreativeWork(is_deleted=True)` to get a deleted work.

tests/api/test_generated_endpoints.py
chrisseto: Can you also add a test to make sure that deleted works are filtered out of list endpoints?

api/views/share.py
```diff
@@ -19,13 +22,35 @@
 class ShareObjectViewSet(viewsets.ReadOnlyModelViewSet):
     ordering = ('-id', )
     pagination_class = CursorPagination
+    permission_classes = [IsDeletedPremissions, ]
```
chrisseto: nitpick: Rather than using lists here, I prefer to use tuples. `(item1, )` or `(item1, item2, ...)`

Lists are mutable objects, someone could accidentally call `ShareObjectViewSet.permission_classes.append(...)` anywhere in the code, leading to unexpected errors.

If you're attribute should not be mutable, like `permission_classes` or `ordering`,  tuples give some peace of mind and indicate that the attribute should not be changed.

api/permissions.py
```diff
@@ -27,3 +30,13 @@ def has_permission(self, request, view):
...
                        'class to be used.')
+
+
+class IsDeletedPremissions(BasePermission):
+    """
+    Permission check for deleted objects.
+    """
+    def has_object_permission(self, request, view, obj):
+        if hasattr(obj, 'is_deleted') and obj.is_deleted:
+                raise PermissionDenied('Query is forbidden for the given object.')
```
chrisseto: This line is double indented.

chrisseto: `Query is forbidden` is a bit generic. Go for something with a bit more detail, like `'This {} has been removed.'.format(obj._meta.verbose_name)`. Removed seems a bit softer than deleted.

tests/api/test_generated_endpoints.py
```diff
@@ -94,6 +95,45 @@ def test_get_data(self, generator, model, route, controlled_values, client, Grap
...
+            '@type': 'dataset',
+        }])
+
+        graph.process()
+        ChangeSet.objects.from_graph(graph, normalized_data_id).accept()
+        preprint.refresh_from_db()
+        assert preprint.is_deleted is True
+
+        response = client.get('/api/v2/{}/{}/'.format('datasets', encoded_id))
+        assert response.status_code == 403
```
chrisseto: You should also check the response body here to ensure that the error message is being sent back correctly and that you're encountering the correct error in the first place.

PR #643, opened 2017-05-02
[SHARE-771][Feature] Elasticsearch janitor task

laurenbarker:
## Purpose 
Find discrepancies between postgres and elasticsearch numbers and reindex missing works.

## Changes
* add elasticsearch `JanitorTask` which finds differences between the database and elasticsearch index and reindexes the missing date ranges in chunks of 500
* use `date_modified` of most recent record in elasticsearch instead of `last_run` for the elastic bot
* add index to `date_created` field
new_indexes:
```
share_agent_date_created_f860e681_uniq
share_agentrelation_date_created_141441e8_uniq
share_agentrelationversion_date_created_2776048c_uniq
share_agentversion_date_created_1134688b_uniq
share_agentworkrelation_date_created_6a125e81_uniq
share_agentworkrelationversion_date_created_24466587_uniq
share_creativework_date_created_d748cb7b_uniq
share_creativeworkversion_date_created_b43cc0f8_uniq
share_workrelation_date_created_76d78396_uniq
share_workrelationversion_date_created_48778122_uniq
share_agentidentifier_date_created_8acb014c_uniq
share_agentidentifierversion_date_created_44a7f2c7_uniq 
share_award_date_created_0d631d11_uniq
share_awardversion_date_created_f5294f8e_uniq 
share_extradata_date_created_22f53cc9_uniq 
share_extradataversion_date_created_dc5dfa07_uniq 
share_tag_date_created_7af766a9_uniq
share_tagversion_date_created_d6dba797_uniq 
share_throughawards_date_created_d1c89159_uniq 
share_throughawardsversion_date_created_fa677f97_uniq 
share_throughcontributor_date_created_6bc1e9a1_uniq 
share_throughcontributorversion_date_created_e5bae6b8_uniq 
share_throughsubjects_date_created_2d723387_uniq 
share_throughsubjectsversion_date_created_02fd8558_uniq 
share_throughtags_date_created_7bf6bc7e_uniq 
share_throughtagsversion_date_created_cbb82dff_uniq 
share_workidentifier_date_created_1ef0c3e4_uniq 
share_workidentifierversion_date_created_23c8c26f_uniq
```


bots/elasticsearch/bot.py
```diff
@@ -160,6 +162,12 @@ def __init__(self, config, started_by, last_run=None, **kwargs):
 
         self.es_client = Elasticsearch(self.es_url)
 
+    def get_most_recently_modified(self):
```
chrisseto: Can you use the elasticsearch client to get this information?

laurenbarker: Yup! (I just figured out how to actually use the client 30 seconds ago)

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +136,73 @@ def serialize(self, source):
...
+        If counts differ, split the date range in half and check counts in smaller ranges
+        Pseudo binary because it can't throw away half the results based on the middle value
+        '''
+        from bots.elasticsearch.bot import ElasticSearchBot
+
+        counts_match, db_count, es_count = self.check_counts_in_range(min_date, max_date)
+
+        if counts_match:
+            return
+        if db_count <= 500:
```
chrisseto: Can the 500 be made into a constant?

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +136,73 @@ def serialize(self, source):
...
+            logger.info('Counts for {} to {} do not match. {} creativeworks in ES, {} creativeworks in database.'.format(min_date.format('%B %-d, %Y %I:%M:%S %p'), max_date.format('%B %-d, %Y %I:%M:%S %p'), es_count, db_count))
```
chrisseto: Please use `%` formats for logging messages. It saves a bit of work if the logs aren't being rendered somewhere and will catch any formatting errors if there happen to be any.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +136,73 @@ def serialize(self, source):
...
+
+        for key, value in self.get_date_range_parts(min_date, max_date).items():
+            self.pseudo_bisection_method(value['min_date'], value['max_date'])
+
+    def do_run(self, es_url=None, es_index=None):
+        self.es_client = Elasticsearch(es_url or settings.ELASTICSEARCH_URL, retry_on_timeout=True, timeout=settings.ELASTICSEARCH_TIMEOUT)
+        self.es_index = es_index
+
+        # get range of date_created in database
+        postgres_high_low = CreativeWork.objects.all().aggregate(lowest=Min('date_created'), highest=Max('date_created'))
```
chrisseto: It's probably safe to assume that the max `date_created` is the current date to save a little bit of time and work.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +136,73 @@ def serialize(self, source):
...
+
+            bot = apps.get_app_config('elasticsearch').get_bot(
+                ShareUser.objects.get(username=settings.APPLICATION_USERNAME),
+                es_filter={'date_created__range': [min_date.isoformat(), max_date.isoformat()]},
+            )
+            bot.run()
+            return
+
+        for key, value in self.get_date_range_parts(min_date, max_date).items():
+            self.pseudo_bisection_method(value['min_date'], value['max_date'])
```
chrisseto: Could you update this to spin these off into celery tasks? Theses tasks are idempotent and will go much faster if we abuse all the computing power available to us.

bots/elasticsearch/tasks.py
chrisseto: Should this be `math.abs(db_count - es_count)`?

Hmmm, now that I think about it we may want to make this logic a bit more complex.

I had this backwards when I originally told you this. We should have the cut off be a timedelta to minimize the amount of data we shove back into elastic.

I think it should be more along the lines of:
If the date range you're looking at is less than 12 hours or the total percentage of missing documents is greater than or equal to 60%.

bots/elasticsearch/bot.py
```diff
@@ -160,6 +161,14 @@ def __init__(self, config, started_by, last_run=None, **kwargs):
...
+        resp = self.es_client.search(
+            index=(self.es_index or settings.ELASTICSEARCH_INDEX),
+            doc_type='creativeworks',
+            body='{"sort": {"date_modified": "desc"}}'
```
laurenbarker: I hadn't considered a max aggregation but it looks like sorting and getting the first result is a little bit faster than. Test results:
* Max aggregation: `'took': 389`
* Sorting: `'took': 163`

bots/elasticsearch/tasks.py
laurenbarker: After discussing, decided on 70% or 500 docs since hours don't have a predictable relationship on the number of documents within that range.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +138,79 @@ def serialize(self, source):
...
+def get_date_range_parts(min_date, max_date):
+    middle_date = min_date.average(max_date)
+    return {
+        'first_half': {'min_date': min_date, 'max_date': middle_date},
+        'second_half': {'min_date': middle_date, 'max_date': max_date}
+    }
+
+
+@celery.task
+def pseudo_bisection_method(self, min_date, max_date):
```
laurenbarker: @chrisseto, is this what you had in mind? I figured the task doesn't need to be logged since there could be a lot of tasks and each one doesn't really mean anything. 

If this kind of task is fine, should it go inside or outside the JanitorTask class? `self` still needs to be passed along to preserve the elastic client but it seems weird to me.

bots/elasticsearch/bot.py
aaxelb: You could even put `'size': 1` in there to eke out a few more ms.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +138,79 @@ def serialize(self, source):
...
+        return
+    if db_count <= MAX_DB_COUNT or 1 - abs(es_count / db_count) >= MIN_MISSING_RATIO:
+        logger.info('Counts for %s to %s do not match. %s creativeworks in ES, %s creativeworks in database.', min_date.format('%B %-d, %Y %I:%M:%S %p'), max_date.format('%B %-d, %Y %I:%M:%S %p'), es_count, db_count)
+        logger.info('Reindexing records created from %s to %s.', min_date.format('%B %-d, %Y %I:%M:%S %p'), max_date.format('%B %-d, %Y %I:%M:%S %p'))
+
+        bot = apps.get_app_config('elasticsearch').get_bot(
+            ShareUser.objects.get(username=settings.APPLICATION_USERNAME),
+            es_filter={'date_created__range': [min_date.isoformat(), max_date.isoformat()]},
+        )
+        bot.run()
```
aaxelb: Should the task also remove records from the index which don't match a work in the database? I'm not sure if that's ever been a problem, but it seems possible.

bots/elasticsearch/bot.py
laurenbarker: Oh shoot, I did that in my test which is probably why it was a little faster. Good catch ⚾️

aaxelb: Out of curiosity, did you use `'size': 0` in the max aggregation test?

laurenbarker: Yup! I ran it a few times and the time varied but was always slightly slower.

aaxelb: Cool cool 👍

bots/elasticsearch/tasks.py
laurenbarker: Seems like a good idea. Created [SHARE-841](https://openscience.atlassian.net/browse/SHARE-841) for the JanitorTask to remove records from the index and also log which records are missing.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +137,77 @@ def serialize(self, source):
...
+        date_created__range=[min_date, max_date]
+    ).count()
+    partial_es_count = self.es_client.count(
+        index=(self.es_index or settings.ELASTICSEARCH_INDEX),
+        doc_type='creativeworks',
+        body={
+            'query': {
+                'range': {
+                    'date_modified': {
+                        'gte': '{}'.format(min_date.isoformat()),
```
chrisseto: Nit-pick: Just do `min_date.isoformat()`. The format does nothing.

bots/elasticsearch/tasks.py
chrisseto: >  I figured the task doesn't need to be logged since there could be a lot of tasks and each one doesn't really mean anything.

Agreed

> self still needs to be passed along to preserve the elastic client but it seems weird to me.

`self` won't be JSON serializable.

You can reconstruct the Elasticsearch client and accept all the same options (`ES_URL`, `ES_INDEX`, etc)
Or build this logic into the main task.

I don't really have a preference, follow your ❤️

chrisseto: Nit-pick: Can you remove `_method` from the name? Technically it's not a method and otherwise would be redundant.

bots/elasticsearch/tasks.py
```diff
@@ -132,3 +137,77 @@ def serialize(self, source):
...
+    '''
+    def do_run(self, es_url=None, es_index=None):
+        self.es_client = Elasticsearch(es_url or settings.ELASTICSEARCH_URL, retry_on_timeout=True, timeout=settings.ELASTICSEARCH_TIMEOUT)
+        self.es_index = es_index
+
+        # get range of date_created in database; assumes current time is the max
+        min_date = pendulum.instance(CreativeWork.objects.all().aggregate(Min('date_created'))['date_created__min'])
+        max_date = pendulum.utcnow()
+
+        pseudo_bisection_method.apply_async((self, min_date, max_date))
```
chrisseto: Do pendulum objects get serialized correctly?

share/models/base.py
```diff
@@ -52,7 +52,7 @@ class ShareObjectMeta(ModelBase):
     share_attrs = {
         'change': lambda: models.OneToOneField(Change, related_name='affected_%(class)s', editable=False, on_delete=DATABASE_CASCADE),
         'date_modified': lambda: models.DateTimeField(auto_now=True, editable=False, db_index=True, help_text=_('The date this record was modified by SHARE.')),
-        'date_created': lambda: models.DateTimeField(auto_now_add=True, editable=False, help_text=_('The date of ingress to SHARE.')),
+        'date_created': lambda: models.DateTimeField(auto_now_add=True, editable=False, db_index=True, help_text=_('The date of ingress to SHARE.')),
```
chrisseto: Once we merge https://github.com/CenterForOpenScience/SHARE/pull/645 could you use it to add the new index?

bots/elasticsearch/tasks.py
laurenbarker: Oh yeah, "method" was a reference to the [Bisection Method](https://en.wikipedia.org/wiki/Bisection_method) in math but I see how that would be confusing.

share/models/base.py
laurenbarker: Yup!

chrisseto:
#645 has been merged.

share/migrations/0032_auto_20170502_1816.py
```diff
@@ -0,0 +1,155 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.7 on 2017-05-02 18:16
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
laurenbarker: @chrisseto, the migration from #645 replaces this migration, correct?

chrisseto: Correct

aaxelb:
Should there be a check that only one elastic task is running at a time? (Maybe janitor task, too, though it probably doesn't run often enough to matter.) If they overlap they could index the same things twice, which isn't a problem but could slow things down. Is there any risk of weirder behavior from the bots jumping over each other?

laurenbarker:
@chrisseto, fixed the JSON serializable issue 💵

share/models/base.py
chrisseto: `db_index` can be removed

PR #644, opened 2017-05-04
[SHARE-839][Feature] Allow creativeworks to be changed via the admin …

chrisseto:
…interface


share/admin/readonly.py
```diff
@@ -0,0 +1,43 @@
...
+class SetOfEverything(list):
+    def __contains__(self, other):
+        return True
+
+
+## adopted from http://mike.hostetlerhome.com/blog/2012/11/13/add-a-read-only-role-to-django-admin/
+class ReadOnlyAdmin(admin.ModelAdmin):
+
+    def has_add_permission(self, request, obj=None):
+        return self.__user_is_readonly(request)
```
aaxelb: If I'm reading this right, only read-only users have add/delete permission... is that the intent?

share/admin/share_objects.py
```diff
@@ -0,0 +1,148 @@
...
+    # ordering = ('-date_modified', )
+    readonly_fields = ('date_created', 'date_modified', )
+    show_full_result_count = False
+    paginator = FuzzyPaginator
+
+    def get_search_results(self, request, queryset, search_term):
+        ret = super().get_search_results(request, queryset, search_term)
+        return ret
+
+    def get_type(self, obj):
```
aaxelb: This method name makes the column have the title "GET TYPE", which is kinda weird. But also doesn't particularly matter.

share/admin/share_objects.py
```diff
@@ -0,0 +1,148 @@
...
+    list_display = ('encoded_id', 'id', 'get_type', 'title', 'date_modified', 'date_created', )
+    search_fields = ('identifiers__uri', )
+    show_change_link = False
+
+    def get_search_results(self, request, queryset, search_term):
+        # Overriden because there is no way to opt out of a case insensitive search
+        search_fields = self.get_search_fields(request)
+        use_distinct = bool(search_term)
+        if search_fields and search_term:
+            orm_lookups = ['{}__startswith'.format(search_field) for search_field in search_fields]
```
aaxelb: Is `_contains` too slow to use here? If you're giving a partial identifier, it's probably the end bit, like a DOI or something.

Also, I think it'd be nice to handle searching for a SHARE ID here.

chrisseto: `__contains` is significantly slower, the queries take > 60secs on prod (Way more).
The index on `uri` can only be used for `__startswith` or an exact match.

share/admin/__init__.py
```diff
@@ -21,18 +21,23 @@
...
+from share.models.banner import SiteBanner
 from share.models.celery import CeleryTask
 from share.models.change import ChangeSet
 from share.models.core import NormalizedData, ShareUser
+from share.models.creative import AbstractCreativeWork
 from share.models.ingest import RawDatum, Source, SourceConfig, Harvester, Transformer
 from share.models.logs import HarvestLog
 from share.models.registration import ProviderRegistration
-from share.models.banner import SiteBanner
-from share.readonlyadmin import ReadOnlyAdmin
```
chrisseto: `SiteBanner` got sorted to the top of the list 👍

aaxelb: I mean [this file](https://github.com/CenterForOpenScience/SHARE/blob/ece47d223b5cee500e48884af645e0315ec41fe4/share/readonlyadmin.py) should be deleted, because it was moved to/replaced by `share/admin/readonly.py`

chrisseto: Completely misread that, sorry. Brain turned that into "Looks like only readonlyadmin".
Will remove.

share/admin/share_objects.py
aaxelb: I suspected as much :( and I suppose indexing for a full-text search is more trouble than it's worth, here.

Searching by SHARE ID should be easy, though, and probably more useful.

share/admin/readonly.py
chrisseto: I think so. It's pretty hacky, you have to give a user some kind of permission to view objects. In the rest of the code it disallows all action :/

share/admin/share_objects.py
chrisseto: Fixed

share/admin/share_objects.py
chrisseto: Added

share/admin/share_objects.py
```diff
@@ -0,0 +1,159 @@
...
+    inlines = (SourcesInline, )
+    list_filter = (TypedModelFilter, SourcesFilter, )
+    # Django forces order by pk desc which results in a seqscan if we add anything else
+    # ordering = ('-date_modified', )
+    readonly_fields = ('encoded_id', 'date_created', 'date_modified', )
+    show_full_result_count = False
+    paginator = FuzzyPaginator
+
+    # view_on_site doesn't actually call get_absolute_url
+    def view_on_site(self, obj):
```
aaxelb: I think if get_absolute_url gave an absolute url (scheme and all), it'd work without having to specify view_on_site

aaxelb:
Looks good 💾

PR #645, opened 2017-05-05
[SHARE-837][Feature] Add support for CONCURRENT indexes

chrisseto:


db/backends/postgresql/schema.py
```diff
@@ -8,6 +8,8 @@ class DatabaseSchemaEditor(PostgresDatabaseSchemaEditor):
     ON_DELETE_DEFAULT = 'NO ACTION'
     ON_UPDATE_DEFAULT = 'NO ACTION'
 
+    sql_create_index_concurrently = 'CREATE INDEX CONCURRENTLY %(name)s ON %(table)s (%(columns)s)%(extra)s'
```
aaxelb: Just in case we ever want a concurrent GIN index or something, this should probably preserve the `%(using)s` in `sql_create_index`:
```
"CREATE INDEX CONCURRENTLY %(name)s ON %(table)s%(using)s (%(columns)s)%(extra)s"
```

share/models/logs.py
```diff
@@ -214,6 +215,9 @@ class HarvestLog(AbstractBaseLog):
 
     class Meta:
         unique_together = ('source_config', 'start_date', 'end_date', 'harvester_version', 'source_config_version', )
+        indexes = [
+            ConcurrentIndex(fields=['harvester_version'])
```
aaxelb: Should there be a migration for this, or are you saving that for another (janitor) PR?

chrisseto: Ah, that was a mistake. I'll remove it.
This is mainly for the janitor PR.

PR #650, opened 2017-05-08
[SHARE-633][Fix] Parse datacite description and tags

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-633

## Problem

This is a potential problem with every record that we are getting from Datacite because of an issue with the datacite transformer.

## Solution:

- Update the datacite source.yaml file.
- Update the datacite transformer.


share/transformers/org_datacite.py
```diff
@@ -490,7 +494,7 @@ def get_schema(self, type):
     )
     description = tools.RunPython(
         force_text,
-        tools.Try(ctx.record.metadata['oai_datacite'].payload.resource.descriptions.description[0])
+        tools.Try(ctx.record.metadata['oai_datacite'].payload.resource.descriptions)
```
chrisseto: You might want to use `tools.Join` here. If there are multiple descriptions it will join them all into a single string.

share/transformers/org_datacite.py
```diff
@@ -551,12 +556,12 @@ def get_schema(self, type):
     tags = tools.Map(
         tools.Delegate(ThroughTags),
         tools.RunPython(
-            force_text,
+            'text_list',
             tools.Concat(
                 tools.Maybe(tools.Maybe(ctx.record, 'metadata')['oai_datacite'], 'type'),
                 tools.RunPython(
                     'text_list',
-                    (tools.Concat(tools.Try(ctx.record.metadata['oai_datacite'].payload.resource.subjects.subject)))
```
laurenbarker: If `Concat` was the problem for tags, it should be removed for the subjects as well (https://github.com/CenterForOpenScience/SHARE/pull/650/files#diff-bedec94698aaa5a42507c84db3ae44d1R549).

laurenbarker:
https://openscience.atlassian.net/browse/SHARE-848 has been added to address the broader concerns. As soon as the subjects field is addressed this should be good to go  🎉

sheriefvt:
@laurenbarker thanks for your feedback. I updated the subject field as requested. 🐼

PR #653, opened 2017-05-09
[SHARE-822][Task] Proposal for supporting custom taxonomies

chrisseto:


whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,112 @@
...
+   - Migrate the existing Subjecting system to this new solution
+   - Continue to use the Bepress taxonomy as SHARE's central taxonomy
+
+
+## Changes to SHARE:
+  - New Model: "Taxonomy"
+  - Changes to Model: "Subject"
+    - Unique indexes on Subjects will move from `[name]` to `[name, taxonomy_id]` if the OSF requirements are ammenable to this
+      Otherwise it will become `[name, taxonomy_id, parent_id]`
+  - If we can upgrade to elasticsearch 5.4+ in time SHARE will index subjects using the new Path Hierarchy Tokenizer.
```
aaxelb: [2.x has that, too](https://www.elastic.co/guide/en/elasticsearch/reference/2.3/analysis-pathhierarchy-tokenizer.html). We should still upgrade to 5.x, though.

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,112 @@
...
+
+## Subject
+
+| Column        |   Type   | Indexed | Nullable | FK  | Default | Description                                                                                        |
+| :------------ | :------: | :-----: | :------: | :-: | :-----: | :------------------------------------------------------------------------------------------------- |
+| name          |   text   |         |          |     |         | The subject name                                                                                   |
+| deleted       |   bool   |         |          |     |         |                                                                                                    |
+| handle        |   text   | unique  |          |     |         | A unique refence to this subject that Sources may use to specify which subject needs to be altered |
+| taxonomy_id   |   int    |    ✓    |          |  ✓  |         | The taxonomy this subject belongs to                                                               |
+| alias_id      |   int    |    ✓    |          |  ✓  |         | The Bepress equivalent of this subject. Set to id if this subject belongs to Bepress.              |
```
chrisseto: I'm a bit skeptical that we'll be using bepress forever, so I was going for something non-specific. It is fairly confusing.

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,112 @@
...
+  Subject Echo -> Bepress D
+
+Given SHARE is aware of custom Taxonomy X
+When SHARE recieves a work, "Gamma", with subjects [Alpha, Bravo, Charlie]
+Then SHARE will remove all parent subjects as they will be implied by subject "Charlie"
+
+  Reasoning:
+    If at any point "Charlie" is moved to another location in Taxonomy X no changes need to
+    be made to "Gamma", we simply find all works that haven't been modified since "Charlie"
+    has been modified and re-index them.
```
laurenbarker: When SHARE receives a work "Gamma", with subjects  [Alpha, Bravo, Charlie] it makes sense to assume that "Charlie" is the MVS (most valuable subject 🥇 ) since it is the most specific. Since "Charlie" is the MVS, if the taxonomy changes, it would make sense to assume the parent subjects based on "Charlie". If "Gamma" should also have the subject "Bravo", then "Gamma" would need to be updated on the source side to reflect that change. In that case we should receive an update for "Gamma" from the source.

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,113 @@
...
+
+  Reasoning:
+    "Delta" will have to exist for SHARE to detect the need to remove the subject from existing
+    works.
+ ```
+### Strange Situations:
+  - Custom Taxonomy item added
+    - Foxtrot is added to Taxonomy X
+  - Custom Taxonomy item removed
+    - Bravo is removed to Taxonomy X
```
laurenbarker: removed from

whitepapers/tickets/SHARE-822.md
laurenbarker: I think not using `bepress_id` is a good idea although maybe not `alias_id`. Not that I have a better idea... maybe `default_id` or `master_id`

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,113 @@
...
+    - Echo is removed to Taxonomy X
+  - Custom Taxonomy item renamed
+    - Echo -> Romeo is added to Taxonomy X
+  - Custom Taxonomy item redefined
+    - Alpha -> Bepress B is added to Taxonomy X
+  - Custom Taxonomy item duplicated
+    - Foxtrot -> Bepress A is added to Taxonomy X
+  - Preprint with Custom Taxonomy moves providers
```
laurenbarker: I think this would be another situation where it would be updated on the source.

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,112 @@
...
+    be made to "Gamma", we simply find all works that haven't been modified since "Charlie"
+    has been modified and re-index them.
+
+Given SHARE is aware of custom Taxonomy X
+When SHARE recieves a work, "Gamma", with subjects [Charlie, Bepress C, Bepress D]
+Then SHARE will not modify the submitted subjects
+
+  Reasoning:
+    "Charlie" may currently imply that "Bepress C" is attached to "Gamma". If "Charlie" is
+    ever redefined to be a different Bepress subject that implication will no longer be true.
```
laurenbarker: Source should provide an update in this situation as well.

whitepapers/tickets/SHARE-822.md
aaxelb: I guess that makes sense. The only weird bit might be when nothing is actually changed on the source's side (e.g. there was an error in the taxonomy we were given), but we require them to mark it updated so we'll reharvest it. For now, I think that's fine. But if/when it's super easy for us to flag a SUID for reingestion and have a janitor or something pick it up, that seems like the ideal solution.

whitepapers/tickets/SHARE-822.md
aaxelb: What are we calling the role of the bepress taxonomy? The central/main/master/canonical/SHARE/official taxonomy? I think this field should reflect the language we use elsewhere, like `central_synonym_id` or `central_alias_id`. I kinda like "canonical" but feel like the word might be ruined at this point...

chrisseto: `central_synonym_id` + 1

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,232 @@
...
+    "@id": {"type": "string"},  // Unchanged
+    "@type": {"enum": ["subject"] },  // Unchanged
+    "name": {"type": "string"}, // Unchanged
+    "parent": {"type": {"$ref": "#/definitions/reference"}},  // Unchanged
+
+    "central_synonyms": {
+      "type": {"$ref": "#/definitions/reference"}
+      "default": null,
+      "description": "The central (Currently Bepress) subject that this subject is equivalent to."
+      // Nullable right now. It might make sense to make this a self reference for Bepress Subject
```
mfraezz: We should probably decide to either make `null` semantically significant going forward, or be self-referential from the get-go.

chrisseto: I lean towards self-referential as it feels a bit less error prone.

whitepapers/tickets/SHARE-822.md
mfraezz: I'd vote `central_alias_id`, but only because `alias` is the term used to describe this relationship on OSF. Either's fine

whitepapers/tickets/SHARE-822.md
```diff
@@ -0,0 +1,232 @@
...
+```javascript
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "type": "object",
+
+  "properties": {
+    "@id": {"type": "string"},  // Unchanged
+    "@type": {"enum": ["subject"] },  // Unchanged
+    "name": {"type": "string"}, // Unchanged
+    "parent": {"type": {"$ref": "#/definitions/reference"}},  // Unchanged
```
mfraezz: I could be wrong about this, but it [doesn't seem like](https://github.com/CenterForOpenScience/osf.io/blob/develop/website/preprints/tasks.py#L72) this is unchanged.

PR #655, opened 2017-05-12
[SHARE-315][Feature] Mendeley Data Source

laurenbarker:
Requires credentials that can be obtained from http://dev.mendeley.com/myapps.html.
If anyone wants to test locally I can keybase you my credentials.

## Purpose
Add Mendeley Data as a source.

## Changes
Added harvester and transformer for Mendeley Data.

## Side Effects
None

## Ticket
https://openscience.atlassian.net/browse/SHARE-315


share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+                    for contributor in dataset['contributors']:
+                        try:
+                            profile_resp = self.get_contributor_profile(headers, contributor['profile_id'])
+                            contributor['full_profile'] = profile_resp.json()
+                        except KeyError:
+                            continue
+                yield (dataset['id'], dataset)
+
+            if 'Link' in resp.headers:
+                next_link = resp.headers['Link'].split(';')
```
chrisseto: Looks like requests might do this to?
http://docs.python-requests.org/en/master/user/advanced/?highlight=Link#link-headers

share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+                resp = self.requests.get(next_link[0].strip('<>'), headers=headers)
+            else:
+                break
+
+    def get_contributor_profile(self, headers, contributor_uuid):
+        token = headers['Authorization']
+        contributor_headers = {
```
chrisseto: Might be worth it to just add these to the harvesters session?

share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+            'Authorization': token,
+            'Accept': 'application/vnd.mendeley-profiles.1+json'
+        }
+        profile_url = furl('https://api.mendeley.com/profiles/').join(contributor_uuid).url
```
chrisseto: Could you make this a constant?

share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+                next_link = resp.headers['Link'].split(';')
+                resp = self.requests.get(next_link[0].strip('<>'), headers=headers)
+            else:
+                break
+
+    def get_contributor_profile(self, headers, contributor_uuid):
+        token = headers['Authorization']
+        contributor_headers = {
+            'Authorization': token,
+            'Accept': 'application/vnd.mendeley-profiles.1+json'
```
chrisseto: Also should be a constant.

share/transformers/com_mendeley_data.py
```diff
@@ -0,0 +1,435 @@
...
+    'related_to': '',
+    'derived_from': 'IsDerivedFrom',
+    'source_of': 'IsDerivedFrom',
+    'compiles': 'Compiles',
+    'compiled_by': 'Compiles',
+    'cites': 'Cites',
+    'cited_by': 'Cites',
+}
+
+INVERSE_RELATIONS = (
```
chrisseto: Nit-pick: If you're just using these to check membership a `set` is much more efficient than a `tuple`

share/transformers/com_mendeley_data.py
```diff
@@ -0,0 +1,435 @@
...
+        if inverse and relation in INVERSE_RELATIONS:
+            results.append(option)
+        elif not inverse and relation in RELATIONS:
+            results.append(option)
+    return results
+
+
+def get_relation_type(relation_type):
+    normalized_relation = RELATION_MAP[relation_type]
+    return normalized_relation or 'WorkRelation'
```
chrisseto: `.get(relation_type, 'WorkRelation')`?

share/transformers/com_mendeley_data.py
```diff
@@ -0,0 +1,435 @@
...
+
+
+class WorkIdentifier(Parser):
+    uri = ctx
```
chrisseto: IRI parser? Is that a thing that we still have to do?

laurenbarker: It is. I put the IRI parser in the [creativework](https://github.com/CenterForOpenScience/SHARE/pull/655/files#diff-fbb6801351f10ca0870441afa2554c4aR388) parser because if it gets to the the workidentifier and then fails, the whole record fails validation.

share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+        while True:
+            for dataset in resp.json()['results']:
+                # modified_since filters on publish_date
+                if pendulum.parse(dataset['publish_date']) > end_date:
+                    break
+                # Send another request to get useful contributor information
+                if 'contributors' in dataset:
+                    for contributor in dataset['contributors']:
+                        try:
+                            profile_resp = self.get_contributor_profile(headers, contributor['profile_id'])
```
laurenbarker: Caching would only provide value if the same contributors are appearing in multiple records within the same harvest, correct? I'll check and see if that's common.

share/harvesters/com_mendeley_data.py
```diff
@@ -0,0 +1,82 @@
...
+class MendeleyHarvester(BaseHarvester):
+    VERSION = 1
+
+    def get_token(self):
+        """ Mendeley gives tokens that last for one hour. A new token will be
+        requested everytime the harvester is run to ensure the access token is
+        valid.
+        """
+        url = 'https://api.mendeley.com/oauth/token'
+        username_password = settings.MENDELEY_API_CLIENT_ID + ':' + settings.MENDELEY_API_CLIENT_SECRET
```
laurenbarker: I'm not following. You want the client ID and the client secret to be harvester kwargs and not environment variables?

share/harvesters/com_mendeley_data.py
laurenbarker: There appear to be few, if any, duplicate contributors in one harvest so I don't think it would be worth caching the responses. What do you think?

chrisseto: Not worth it, leave it as is 👍

chrisseto:
Looks good to go, just two final things:
1. The ticket said their requested rate limit was 1 request a minute. Did that change or should this be updated?
2. Do you want to add the encrypted field as part of this PR or should we add that in later so we can send this out ASAP?

laurenbarker:
@chrisseto
1. Updated the rate limit. I couldn't find anything in their docs about a rate limit so 1 request per minute seems good. 
2. The encrypted field should probably go in a separate PR. Especially since Mendeley has been waiting on this harvester for like a month.

chrisseto:
Sounds good, I'll put this on staging tomorrow morning 👍

PR #659, opened 2017-05-17
[SHARE-848][Feature] Move 'force_text' to utils

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-848

## Problem:
Right now, the transformer for datacite, oai, mods, and clinicaltrial all use some variation of force_text.

## Solution:
* Move force_text from oai, clinicaltrials, mods, and datacite to utils
* Use one version for all sources
* Remove text_list and replace with force_text
* Add tests to force_text


share/transform/chain/utils.py
```diff
@@ -36,3 +41,34 @@ def format_address(address1='', address2='', city='', state_or_province='', post
         return '{}, {}'.format(city, state_or_province)
 
     return address1
+
+
+def force_text(data):
```
laurenbarker: @sheriefvt, this is so much better 🎉 . Can you add some tests for `force_text`? That way we can be sure that it is getting everything.

sheriefvt: I am working on it now..thanks

chrisseto:
@sheriefvt could you fix the merge conflicts?

PR #660, opened 2017-05-17
[SHARE-739][Improvement] Check quality of OAI sources

laurenbarker:
*I had to run `./manage.py loadsources --overwrite` to update the existing source configs*

## Purpose
Check quality of OAI sources and make sure they all provide links back to the source.

## Changes
* Add comment too source config if DSpace or Digital Commons/Bepress
* Update base urls
* Add `earliest_date` values
* Add task that stores stats for sources
* Display on admin

screen shot 2017-06-13 at 2 07 14 pm


share/sources/edu.oaktrust/source.yaml
```diff
@@ -1,7 +1,8 @@
+# DSpace
 configs:
 - base_url: http://oaktrust.library.tamu.edu/dspace-oai/request
   disabled: false
-  earliest_date: null
+  earliest_date: 2004-11-10 20:56:05  # their earliestDatestamp is most recent
```
aaxelb: lol

tests/share/test_sources.py
```diff
@@ -0,0 +1,87 @@
...
+import pendulum
+import requests
+from requests.exceptions import Timeout
+from sickle import Sickle
+
+from share.models import SourceConfig
+
+
+@pytest.mark.django_db
+class TestSources:
```
aaxelb: Should these tests be disabled by default? I think it's great we have them, but I suspect they take a while and it doesn't seem necessary to run them every time I make a PR.

laurenbarker: Yeah I was thinking that too. Especially since they could fail and have nothing to do with the PR. They should be run fairly frequently though?

chrisseto: It could be a nightly task that we run in prod?

laurenbarker:
I did include the ones that looked they were displaying the published date but could definitely comment those out. The published date will always be before any records though. Thoughts?

### Lastest
*edu.oaktrust.mods*:  2017-05-16T09:19:07
### Error?
*edu.scholarsarchiveosu.mods*:  0011-01-01T00:00:00Z
*edu.uwashington.mods*:  2083-03-01T08:00:00Z
### Published
*gov.nodc*:  1996-10-09
*org.philpapers*:  1990-01-01T00:00:00Z
*org.ttu.mods*:  1989-05-01T05:00:00Z
*es.csic.mods*:  1989-11-28T07:40:02Z
*edu.umich.mods*:  1983-01-01T05:00:00Z
*edu.udc.mods*:  1981-09-01T13:37:51Z
*be.ghent.mods*:  1970-01-01T00:00:01Z
*edu.citeseerx*:  1970-01-01
*br.pcurio*:  1970-01-01
*ch.cern*:  1969-12-31T23:00:00Z
*edu.vtech.mods*:  1900-02-02T05:00:00Z
*edu.icpsr*:  01-01-1900
*pt.rcaap*:  1900-01-01T00:00:00Z
*org.ucescholarship*:  1885-05-01
*com.nature*:  1869-11-04

tests/share/test_sources.py
laurenbarker: Good idea, that would be the best

aaxelb:
If there's an easy way to get something closer to the actual earliest datestamp for the "published" ones, I feel like that'd be nicer. If it'd require wading through their OAI feed until you find it, though, not worth it.

I think those first three should just have `null`, unless you already found a reasonable value (is the 2004 date for oaktrust right?)

laurenbarker:
* There isn't an easy way to get the earliestDatestamp that I know of
* The error ones aren't included
* 2004 is oaktrust  

I'll remove the published dates. We are going to email all of these people and I wanted the test to alert us that they fixed their earliestDatestamp. They will most likely respond to the email though.

share/models/sources.py
```diff
@@ -0,0 +1,40 @@
...
+        return (self.config.label,)
+
+    def __repr__(self):
+        return '<{}({}, {})>'.format(self.__class__.__name__, self.pk, self.config.label)
+
+    def __str__(self):
+        return '{}: {}'.format(self.config.source.long_title, self.config.label)
+
+
+class SourceStatOAI(SourceStat):
```
laurenbarker: :( Yeah...it doesn't make any sense to do it that way. Didn't really think that through...

chrisseto: Pinging @cwisecarver for his opinion.

cwisecarver: Just to codify the conversation that @chrisseto and I had. Multi-table inheritance is basically the root of all evil (in most situations) because it forces an implicit join. I'd say if you're going to use these extra fields more often then not I'd put them on the original model and make them nullable. If you're not going to use them very often or they're going to be duplicated often I'd put them in a separate model and FK from the original model to the new model so they can be re-used. If you need them to be independent the FK should point the other way, which is basically the same thing as multi-table inheritance except that it's not going to do the join implicitly.

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+from lxml import etree
+import pendulum
+
+from share.models import SourceStat
+from share.models import SourceConfig
+
+logger = logging.getLogger(__name__)
+
+
+class SourceStatus():
```
chrisseto: You don't need the `()`s here.

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def assert_no_exception(self, url, timeout=15.0):
+        try:
+            r = requests.get(url, timeout=timeout)
+        # except all exception and log
+        except Exception as e:
+            logger.warning('Exception received from source: %s', e)
+            return (None, e)
+        return (r, None)
```
chrisseto: https://blog.golang.org/error-handling-and-go

share/tasks.py
```diff
@@ -199,3 +201,31 @@ def schedule_harvests(self, *source_config_ids, cutoff=None):
...
+    ).exclude(
+        harvester__key='oai'
+    )
+    for config in non_oai_sourceconfigs.values():
+        get_source_stats.apply_async((False, config['id']))
+
+
+@celery.shared_task(bind=True)
+def get_source_stats(self, oai, config_id):
+    if oai:
```
chrisseto: Would it make more sense to just load the config here and just check if it's oai?

share/tasks.py
```diff
@@ -199,3 +201,31 @@ def schedule_harvests(self, *source_config_ids, cutoff=None):
...
+    oai_sourceconfigs = SourceConfig.objects.filter(
+        disabled=False,
+        base_url__isnull=False,
+        harvester__key='oai'
+    )
+    for config in oai_sourceconfigs.values():
+        get_source_stats.apply_async((True, config['id']))
+
+    non_oai_sourceconfigs = SourceConfig.objects.filter(
+        disabled=False,
```
chrisseto: `source__is_delete=False`?

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def get_source_stats(self, config_id):
+        sourceconfig = SourceConfig.objects.get(pk=config_id)
```
chrisseto: As the entire class is built around the source config, it might be a bit nicer to just pass it into `__init__` and then just reference it as `self.source_config` or whatever name you prefer.

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+
+            admin_note=admin_note,
+        )
+
+
+class OAISourceStatus(SourceStatus):
+
+    NAMESPACES = {
+        'dc': 'http://purl.org/dc/elements/1.1/',
+        'ns0': 'http://www.openarchives.org/OAI/2.0/',
```
chrisseto: Just saying this to say it, don't bother changing anything. At somepoint we should just switch this to `'oai`' or change everything to `'nsX'`

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+        'oai_dc': 'http://www.openarchives.org/OAI/2.0/oai_dc/',
+    }
+
+    def get_field_from_identify(self, response, field):
+        parsed = etree.fromstring(response.content, parser=etree.XMLParser(recover=True))
```
chrisseto: Not sure if there's a great way to do this, might just be worth adding a todo. It would be interesting to see what sources provide invalid XML. IE fail without `recover=True`

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+        return parsed.xpath('//ns0:Identify/ns0:{}'.format(field), namespaces=self.NAMESPACES)[0].text
+
+    def get_source_stats(self, config_id):
+
+        # Known incorrect baseUrl:
+        incorrect_base_urls = {
```
chrisseto: This should be a class level variable.

share/util/source_stat.py
```diff
@@ -0,0 +1,158 @@
...
+            'https://biblio.ugent.be/oai': 'Listed baseURL is their homepage.',
+            'http://purr.purdue.edu/oaipmh': 'Listed baseURL is their homepage.',
+            'https://www.ncbi.nlm.nih.gov/pmc/oai/oai.cgi': 'Listed baseURL is incorrect.',
+            'https://mla.hcommons.org/deposits/oai/': 'isted baseURL is incorrect.',
+            'http://oai.repec.org': 'Listed baseURL redirects.',
+        }
+
+        # Known incorrect earliestDatestamp (all emailed):
+        incorrect_earliest_datestamp = {
```
chrisseto: Also a class level variable.

share/models/sources.py
```diff
@@ -0,0 +1,39 @@
...
+    admin_note = models.TextField(blank=True)
+
+    # OAI specific
+    earliest_datestamp_source = models.DateField(blank=True, null=True)
+    earliest_datestamps_match = models.BooleanField(default=False)
+
+    base_url_source = models.TextField(blank=True, null=True)
+    base_urls_match = models.BooleanField(default=False)
+
+    objects = NaturalKeyManager('config.label')
```
chrisseto: ping @aaxelb, I don't think this is a valid use or use case of `NaturalKeyManager`. I think the value needs to be unique?

share/models/sources.py
```diff
@@ -0,0 +1,39 @@
...
+from django.db import models
+
+from share.models.ingest import SourceConfig
+from share.models.ingest import NaturalKeyManager
+
+logger = logging.getLogger(__name__)
+__all__ = ('SourceStat',)
+
+
+class SourceStat(models.Model):
```
chrisseto: Could you add a `grade` or `ok` field, so we can filter for sources that are having issues?

share/models/sources.py
aaxelb: Yeah, natural keys should be unique. If `config` were a one-to-one field, this would work (except it should be `config__label`). But it looks like there could be many SourceStats per SourceConfig?

share/tasks.py
laurenbarker: Are you saying we should get stats on disabled source configs?

share/util/source_stat.py
laurenbarker: Is there something in particular I should be looking at?

PR #619, opened 2017-04-05
[WIP] Whitepaper cleanup

aaxelb:

PR #661, opened 2017-05-19
[SHARE-706] Proposed graph data structure for the Regulator

aaxelb:
For your consideration:
- A description of what the Regulator will do, and how ([whitepapers/ingest/Regulator.md](https://github.com/CenterForOpenScience/SHARE/pull/661/files#diff-3bcff49c6d8807d57b5e443490007e4a))
- A proposed MutableGraph interface that meets the needs of the Regulator ([whitepapers/ingest/Graph.md](https://github.com/CenterForOpenScience/SHARE/pull/661/files#diff-43b323b8186bff4b96ac7cdf8049ec56))
- A basic implementation of that interface, for playing with ([share/regulate/graph.py](https://github.com/CenterForOpenScience/SHARE/pull/661/files#diff-836fb07a19700e3c152d44e446144303))
- Also includes whitepaper cleanup from #619


aaxelb:
I suspect we'll discuss whether the graph should be aware of/depend on the SHARE schema. The implementation here looks at the Django models to figure out how it should treat different fields, which I think will make Regulator steps more understandable and easier to write than if we used a more general-purpose graph interface. Since we plan to mutate graphs only during ingestion, and shouldn't ever ingest data to an old version of the schema, I don't think depending on the models will be a problem.

whitepapers/ingest/Graph.md
```diff
@@ -0,0 +1,29 @@
+# MutableGraph [NOT IMPLEMENTED]
```
chrisseto: While I hate to bring this up again but why did we decide against networkX? It [implements all the basic methods required, add_edge, remove_edge, add_node, remove_node, etc](https://networkx.readthedocs.io/en/stable/tutorial/tutorial.html#adding-attributes-to-graphs-nodes-and-edges) and it also has a [topological sort built into it](http://networkx.readthedocs.io/en/stable/reference/generated/networkx.algorithms.dag.topological_sort.html?highlight=sort).
To top it all of, [it doesn't provide it's own nodes or edges](https://networkx.readthedocs.io/en/stable/tutorial/tutorial.html#what-to-use-as-nodes-and-edges), we can us whatever objects we'd like.

@laurenbarker would you mind weighing in?

aaxelb: Oh, using our own objects as nodes/edges is pretty neat. We'd just have to decide how to translate our schema to/from a networkx graph (are relations/throughs edges or nodes?). I still worry a general-purpose graph library could make implementing regulator steps unnecessarily complicated, but if that happens I'm sure we could smooth things over with some convenience methods or something.

I don't remember why we decided against it in the first place, though.

laurenbarker: In the context of the other libraries I was looking at, networkX was the slowest and doesn't implement graph filtering. I think my main issue with the library was the lack of graph filtering.  If we choose to use it we would need to augment the functionality a little bit but it should probably be reconsidered for the base of our graph interface. That being said, the additions would look a lot like what Abram's already done here -- the code would just need to be refactored to use networkX.

See https://graph-tool.skewed.de/performance
> NetworkX, on the other hand, comes at a distant third with running times in the order of 20 to 170 times slower than graph-tool. This is mostly due to its pure Python implementation, which is known in general to be substantially slower than C/C++ (see here, here and here for further comparisons).

[My Evaluation](https://gist.github.com/laurenbarker/97d28ff75f15a4b42a2646d8dd957837)

aaxelb: I do rather like the interface I ended up with here, which uses an understanding of the schema to vastly simplify implementing regulator logic. I wouldn't be opposed to refactoring it to use networkx internally, assuming that actually results in something simpler to understand/maintain.

PR #662, opened 2017-05-19
[Fix] Upgrade to ES 5's text and keyword types

chrisseto:
* https://www.elastic.co/blog/strings-are-dead-long-live-strings
  * https://www.elastic.co/guide/en/elasticsearch/reference/current/keyword.html
  * https://www.elastic.co/guide/en/elasticsearch/reference/current/text.html

### DIFF
--- a/bots/elasticsearch/bot.py
+++ b/bots/elasticsearch/bot.py
@@ -66,8 +66,7 @@ class ElasticSearchBot(Bot):
 
     EXACT_FIELD = {
         'exact': {
-            'type': 'string',
-            'index': 'not_analyzed',
+            'type': 'keyword',
             # From Elasticsearch documentation:
             # The value for ignore_above is the character count, but Lucene counts bytes.
             # If you use UTF-8 text with many non-ASCII characters, you may want to set the limit to 32766 / 3 = 10922 since UTF-8 characters may occupy at most 3 bytes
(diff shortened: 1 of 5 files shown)


bots/elasticsearch/bot.py
```diff
@@ -66,7 +66,7 @@ class ElasticSearchBot(Bot):
 
     EXACT_FIELD = {
         'exact': {
-            'type': 'string',
+            'type': 'keyword',
             'index': 'not_analyzed',
```
aaxelb: Remove `'index'`

bots/elasticsearch/bot.py
```diff
@@ -79,67 +79,67 @@ class ElasticSearchBot(Bot):
...
-                'id': {'type': 'string', 'index': 'not_analyzed', 'include_in_all': False},
-                'identifiers': {'type': 'string', 'fields': EXACT_FIELD},
-                'justification': {'type': 'string', 'include_in_all': False},
-                'language': {'type': 'string', 'index': 'not_analyzed', 'include_in_all': False},
-                'publishers': {'type': 'string', 'fields': EXACT_FIELD},
-                'registration_type': {'type': 'string', 'index': 'not_analyzed', 'include_in_all': False},
+                'description': {'type': 'text'},
+                'funders': {'type': 'text', 'fields': EXACT_FIELD},
+                'hosts': {'type': 'text', 'fields': EXACT_FIELD},
+                'id': {'type': 'text', 'index': 'not_analyzed', 'include_in_all': False},
```
aaxelb: Should be `'keyword'`. Also, remove `'index'`, or set it to `false` if we don't want works searchable by id

bots/elasticsearch/bot.py
laurenbarker: > index
Should the field be searchable? Accepts true (default) or false.

bots/elasticsearch/bot.py
laurenbarker: Should be `keyword` and remove `'index': 'not_analyzed'`

> index
Should the field be searchable? Accepts true (default) or false.

bots/elasticsearch/bot.py
```diff
@@ -79,67 +79,67 @@ class ElasticSearchBot(Bot):
...
+                'tags': {'type': 'text', 'fields': EXACT_FIELD},
+                'title': {'type': 'text', 'fields': EXACT_FIELD},
+                'type': {'type': 'keyword', 'include_in_all': False},
+                'types': {'type': 'keyword', 'include_in_all': False},
                 'withdrawn': {'type': 'boolean', 'include_in_all': False},
                 'lists': {'type': 'object', 'dynamic': True, 'include_in_all': False},
             },
             'dynamic_templates': [
-                {'exact_field_on_lists_strings': {'path_match': 'lists.*', 'match_mapping_type': 'string', 'mapping': {'type': 'string', 'fields': EXACT_FIELD}}},
+                {'exact_field_on_lists_strings': {'path_match': 'lists.*', 'match_mapping_type': 'text', 'mapping': {'type': 'text', 'fields': EXACT_FIELD}}},
```
aaxelb: This [should still be `'string'`](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/dynamic-templates.html#match-mapping-type)

laurenbarker:
I'll just delete my duplicate comments :( @chrisseto, tests are failing

bots/elasticsearch/bot.py
laurenbarker: False would definitely break the dashboard

chrisseto:
Updated travis.yaml and tests on the way.

chrisseto:
`EXACT_FIELD` needs to still exist so `field_name.exact` will continue to work as expected

bots/elasticsearch/bot.py
aaxelb: I think fetching a document by ID doesn't actually use this field (`hits[0]._id` vs `hits[0]._source.id`), but I agree we should leave it searchable.

aaxelb:
We'd also have to go back to dynamic mappings to get the default `foo.keyword` fields, which would be slightly sad. And I'm not sure whether lowering `ignore_above` from 10922 to 256 could break anything?

laurenbarker:
Ah, didn't know we weren't using dynamic mapping

chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/commit/936369959a589d64f6e1542a44b9f601d0f7770d

PR #663, opened 2017-05-22
[Enhancement] Add user agent to harvester requests

laurenbarker:


project/settings.py
```diff
@@ -472,6 +472,7 @@
 EMBER_SHARE_URL = os.environ.get('EMBER_SHARE_URL', 'http://localhost:4200').rstrip('/') + '/'
 SHARE_API_URL = os.environ.get('SHARE_API_URL', 'http://localhost:8000').rstrip('/') + '/'
 SHARE_WEB_URL = os.environ.get('SHARE_WEB_URL', SHARE_API_URL + EMBER_SHARE_PREFIX).rstrip('/') + '/'
+SHARE_USER_AGENT = os.environ.get('SHARE_USER_AGENT', 'SHAREbot/' + VERSION + ' +' + SHARE_WEB_URL)
```
chrisseto: Can you use `.format`? And I think there should be `()` around the `+` and url.

```
Googlebot/2.1 (+http://www.google.com/bot.html)
```

laurenbarker: Done 💸

PR #664, opened 2017-05-22
[SHARE-575][Analysis] Partially unknown dates

aaxelb:
https://openscience.atlassian.net/browse/SHARE-575

### DIFF
--- a/whitepapers/Dates.md
+++ b/whitepapers/Dates.md
@@ -1,0 +1,80 @@
+# Partially Unknown Dates
+
+## Overview
+Many works in SHARE have the wrong publication date displayed, because many sources provide only partial publication dates for some works, such as "May 2017" or just "2017".
+When the day of the month is omitted, the current date parsing logic assumes the first day of the month. When the month is omitted, it assumes January.
+
+### Goals
+Allow storing and displaying works with partially-known publication dates.
+
+## Changes
+
+Create a custom `PartialDate` class with:
+* Three main properties:
+    * `year`: integer (required)
+    * `month`: integer (optional)
+    * `day`: integer (optional)
+* Missing/null values indicate an unknown date part
+* Validation that set values are the correct type and correspond to a valid date
+    * If `month` is missing, `day` must also be missing
+* A way to construct a `PartialDate` from a `Date`-like object, string, or 8-digit integer (see below)
+
+### Models
+* Custom Django field `PartialDateField` that inherits from `IntegerField`
+    * Store a `PartialDate` instance as an 8-digit integer in the database, `YYYYMMDD`
+    * If `month` or `day` is null, put `00` in the `MM` or `DD` segment, respectively
+        * "February 26, 2017" stored as `20170226`
+        * "February 2017" stored as `20170200`
+        * "2017" stored as `20170000`
+    * This allows sorting and filtering by a `PartialDateField` in the database
+        * Values with unknown month and day are sorted at the beginning of the year
+        * Values with known month but unknown day are sorted at the beginning of the month
+* Make `AbstractCreativeWork.date_published` a nullable `PartialDateField`
+
+### Migrations
+* Migrate columns on `share_creativework` and `share_creativeworkversion`
+    * Create a new nullable `INTEGER` column `date_published_partial`
+    * Find all special case dates, convert (see below), and copy them over appropriately
+    * Convert all remaining values
+    * Tell Django that `date_published`'s column is `date_published_partial`
+    * Add a new field `_date_published` pointing at the column `date_published`
+        * So we can keep it around for a little while
+        * Drop the original `date_published` column in the next release
+* Convert datetime values into integers, as above, dropping the "time" part of the datetime
+    * If a value is at exactly midnight on the first day of the month, assume unknown day
+        * If the month is January, also assume unknown month
+    * Examples:
+        * `2017-03-20 06:02:11` becomes `20170320`
+        * `2017-03-20 00:00:00` becomes `20170320`
+        * `2017-03-01 06:02:11` becomes `20170301`
+        * `2017-03-01 00:00:00` becomes `20170300`
+        * `2017-01-01 06:02:11` becomes `20170101`
+        * `2017-01-01 00:00:00` becomes `20170000`
+
+### Transformers
+* Add a `PartialDateParser` Transformer link that parses a string into a `PartialDate` instance
+    * Should also handle dates split into an array (e.g. `[2017, 3]`)
+* Use `PartialDateParser` instead of `DateParser` for `CreativeWork.date_published` field in all Transformers
+
+### JSON APIs
+* Serialize `PartialDateField`s in JSON as an array of date parts:
+    * `20171104` serialized as `[2017, 11, 4]`
+    * `20171100` serialized as `[2017, 11]`
+    * `20170000` serialized as `[2017]`
+
+### Search Index
+* Change `date_published` mapping to accept multiple date formats
+    * `'format': 'yyyy-MM-dd||yyyy-MM||yyyy'`
+* Update elasticsearch bot to format `date_published` accordingly
+    * `20170101` indexed as "2017-01-01"
+    * `20170100` indexed as "2017-01"
+    * `20170000` indexed as "2017"
+* Add `date_published.year`, `date_published.month`, `date_published.day` multi-fields
+    * Use a pattern tokenizer to pull the correct part out of the date string
+    * Queryable fields, but won't be included in search result body
+
+### UI
+* Anywhere `date_published` is displayed (search results and detail page), format it to omit unknown parts
+    * `[2017, 07, 23]` displayed as "Jul 23, 2017" or "2017-07-23"
+    * `[2017, 07]` displayed as "Jul 2017" or "2017-07"
+    * `[2017]` displayed as "2017"


whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+        * "February 26, 2016" stored as `20160226`
+        * "February 2016" stored as `20160200`
+        * "2016" stored as `20160000`
+    * This allows sorting and filtering by a `PartialDateField` in the database
+        * Values with unknown month and day are sorted at the beginning of the year
+        * Values with known month but unknown day are sorted at the beginning of the month
+* Make `AbstractCreativeWork.date_published` a nullable `PartialDateField`
+
+### Migrations
+* Change the `date_published` columns on `share_creativework` and `share_creativeworkversion` from `DATETIME` to `INTEGER`
```
chrisseto: * Create a new nullable `INTEGER` column `date_published_partial`
* Find all special case dates, modify, and copy them over appropriately 
* Copy and truncate all remaining values
* Tell Django that `date_published`'s column is `date_published_partial`
* Add a new field `_date_published` pointing at the column `date_published`
    * So we can keep it around for a little while
* Drop the original `date_published` column in the next release

whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+    * Examples:
+        * `2016-03-20 06:02:11` becomes `20160320`
+        * `2016-03-20 00:00:00` becomes `20160320`
+        * `2016-03-01 06:02:11` becomes `20160301`
+        * `2016-03-01 00:00:00` becomes `20160300`
+        * `2016-01-01 06:02:11` becomes `20160101`
+        * `2016-01-01 00:00:00` becomes `20160000`
+
+### Transformers
+* Add a `PartialDateParser` Transformer link that parses a string into a `PartialDate` instance
```
chrisseto: Should there be any special casing about January 1st?

whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+    * Should also handle dates split into an array (e.g. `[2016, 3]`)
+* Use `PartialDateParser` instead of `DateParser` for `CreativeWork.date_published` field in all Transformers
+
+### JSON APIs
+* Serialize `PartialDateField`s in JSON as an array of 3 date parts:
```
chrisseto: Is there any reason to go with this over crossref/citeprocs format?

whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+    * `20160000` serialized as `[2016, null, null]`
+
+### Search Index
+* Change `date_published` mapping to accept multiple date formats
+    * `'format': 'yyyy-MM-dd||yyyy-MM||yyyy'`
+* Update elasticsearch bot to format `date_published` accordingly
+    * `20160101` indexed as "2016-01-01"
+    * `20160100` indexed as "2016-01"
+    * `20160000` indexed as "2016"
+* Add unindexed `date_published.date_parts` field that contains an array of date parts, as in the JSON APIs
```
chrisseto: Would an object bet better here? `date_published.month` and the like? That way we could query for things missing days/months/etc.

whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+* Use `PartialDateParser` instead of `DateParser` for `CreativeWork.date_published` field in all Transformers
+
+### JSON APIs
+* Serialize `PartialDateField`s in JSON as an array of 3 date parts:
+    * `20160101` serialized as `[2016, 1, 1]`
+    * `20160100` serialized as `[2016, 1, null]`
+    * `20160000` serialized as `[2016, null, null]`
+
+### Search Index
+* Change `date_published` mapping to accept multiple date formats
```
chrisseto: Could we and/or would it make sense to make a custom analyzer?

whitepapers/Dates.md
```diff
@@ -0,0 +1,72 @@
...
+Many works in SHARE have the wrong publication date displayed, because many sources provide only partial publication dates for some works, such as "May 2016" or just "2016".
+When the day of the month is omitted, the current date parsing logic assumes the first day of the month. When the month is omitted, it assumes January.
+
+### Goals
+Allow storing and displaying works with partially-known publication dates.
+
+## Changes
+
+Create a custom `PartialDate` class with:
+* Three main properties:
```
chrisseto: Would it make sense to also go ahead and handle dates that have months and days but not years?

aaxelb: I feel like no? I've not seen that from any source (crossref doesn't allow it in their partial dates) and it doesn't seem like a useful value. It wouldn't make sense in a sort or filter on `date_published`.

whitepapers/Dates.md
aaxelb: 👍 Much better details.

whitepapers/Dates.md
aaxelb: I think that special case makes sense only for the migration. If a source actually tells us "2016-01-01", we should trust them. If/when we next reingest everything, the ones that are just "2016" will stay partial dates, and the ones that are more specific should be updated.

whitepapers/Dates.md
aaxelb: The nested array (`[ [2016, 3, 1] ]`) bothers me. Citeproc uses it to support date ranges, like `[ [2016, 3, 1], [2016, 4, 1] ]`, but we don't plan to support ranges, as far as I know. If we ever do, it'd have to break some compatibility anyway, so we could switch then.

As for nulls vs variable-length arrays (`[2016, null, null]` vs citeproc's `[2016]`), I'm ambivalent. Both would be easy to handle and understand, as long as we're consistent. Nulls are more explicit, but uglier. Any preference?

whitepapers/Dates.md
aaxelb: If `date_published` were an object field, we couldn't query by `date_published` directly. We'd have to put the date in a subfield like `date_published.date`, which isn't good. Now that I'm thinking about it, `date_published.date_parts` would only work on an object field, which also isn't good...

The point of the `date_parts` field was to make the partial date available in search results in a format consistent with the rest of the API. `_source.date_published` will have the original string form (e.g. "2017-03"). We could put the parts in another field (`date_published_parts`?), either as an array (for consistence, would probably be unindexed) or an object (indexed for ease of querying).

Another option is to use [multi-fields](https://www.elastic.co/guide/en/elasticsearch/reference/current/multi-fields.html). We could use custom analyzers with a [pattern tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-pattern-tokenizer.html) to pull the year/month/day into `date_published.year`, etc. They would be queryable but wouldn't show up in the search results. If the string form is sufficient to display search results (it could even be shown as-is), this would give the nicest query interface.

Thoughts?

whitepapers/Dates.md
aaxelb: I don't think that's necessary; `date` fields already work with multiple input formats.

One caveat with this approach: It will internally store "2016" as a timestamp equal to "2016-01-01 00:00:00". When sorting by `date_published`, you might get `2016-01, 2016, 2016-01-01, 2016` all in a row. When filtering, you'll only get "2016" results if you include January 1 in your date range. I think these are acceptable quirks.

(The sorting one could be fixed if we add a `date_published.exact` multi-field that contains the original string, and sort by that.)

whitepapers/Dates.md
aaxelb: A benefit of variable-length arrays I just realized: It passively enforces that you can't have a day without a month, or a month without a year. I think I prefer that over nulls now.

whitepapers/Dates.md
laurenbarker: Are we going to write a parser for all common date formats (e.g. `19 Nov, 2014`, `Nov 2012`, `2001-1-01`, `11/20/1990`, etc. )? We won't be able to use pendulum to parse values out since it would assign defaults. Unless we want to first try parsing the date and if it doesn't work try pendulum's parse? We would also need to enforce valid dates, upper bounds, and lower bounds with the `PartialDateParser`.

aaxelb: I think we should write a parser for all the partial date formats we can think of (just year is easy, and there aren't that many ways to write year and month -- we should use dateutil's [parserinfo.month()](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parserinfo) for month names), then fall back to pendulum.

I was thinking the `PartialDate` class itself should enforce valid dates (like valid day/month numbers). Enforcing our 1217-2117 bounds (or whatever they are) seems like a job for the regulator, but for the moment, I agree, `PartialDateParser`.

laurenbarker: [Arrow docs](http://arrow.readthedocs.io/en/latest/) have a nice table of common date formats:

Token | Output
------ | --------
YYYY | 2000, 2001, 2002 ... 2012, 2013
YY | 00, 01, 02 ... 12, 13
DDDD | 001, 002, 003 ... 364, 365
DDD | 1, 2, 3 ... 4, 5

Token | Output
------ | --------
MMMM | January, February, March ...
MMM | Jan, Feb, Mar ...
MM | 01, 02, 03 ... 11, 12
M | 1, 2, 3 ... 11, 12
DD | 01, 02, 03 ... 30, 31
D | 1, 2, 3 ... 30, 31
Do | 1st, 2nd, 3rd ... 30th, 31st

Token | Output
------ | --------
dddd | Monday, Tuesday, Wednesday ...
ddd | Mon, Tue, Wed ...
d | 1, 2, 3 ... 6, 7

whitepapers/Dates.md
aaxelb: @chrisseto I updated this to use the multi-field approach, which seems nicest to me.

PR #667, opened 2017-05-25
[SHARE-777][Feature][HOLD] Add tind.io harvester for AgEcon

aaxelb:
* Remove the `edu.ageconsearch` harvester and transformer
* Add `io.tind` harvester we could use for any source using [Tind IR](http://info.tind.io/IR) (from their press releases, looks like just AgEcon so far)
* Harvest AgEcon in MODS format

Waiting on a fix on filtering by date range.


share/transformers/mods.py
```diff
@@ -240,7 +240,8 @@ class MODSCreativeWork(Parser):
                     lambda obj: 'invalid' not in obj,
                     tools.Concat(
                         tools.Try(ctx['mods:identifier']),
-                        tools.Try(ctx.header['identifier'])
+                        tools.Try(ctx.header['identifier']),
+                        tools.Try(ctx['mods:location']['mods:url']),
```
laurenbarker: @aaxelb, do you want to pull the location url stuff into another PR since this one is on hold and also resolves https://openscience.atlassian.net/browse/SHARE-861?

share/transformers/edu_ageconsearch.py
```diff
@@ -1,138 +0,0 @@
-import re
```
laurenbarker: Do we want to remove the old transformer? If we re-transform the old data we will still need this code.

share/transformers/mods.py
aaxelb: Makes sense, will do.

share/transformers/edu_ageconsearch.py
aaxelb: Hmm, I'd kinda rather it be removed, since we'll never get new data for it, but I didn't think about the old data. I'll put it back in for now.

aaxelb: In the same spirit, I restored the old source config (disabled and with no harvester) so the system knows what to do with the old data.

PR #668, opened 2017-05-25
[SHARE-858][Fix] Browsable API docs not rendering markdown

aaxelb:
The upgraded Django REST Framework has optional dependencies to render markdown. We have now opted for them.

https://openscience.atlassian.net/browse/SHARE-858

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,6 +25,7 @@ gevent==1.1.1  # MIT
 graphene==1.4  # MIT
 jsonschema==2.5.1  # MIT
 lxml==3.6.0  # BSD
+markdown==2.6.8  # BSD
 nameparser==0.4.0  # LGPL
 newrelic==2.68.0.50  # newrelic APM agent, Custom License
 pendulum==0.6.2  # MIT


requirements.txt
```diff
@@ -4,6 +4,7 @@ bleach==1.4.3  # Apache 2.0
 boto3==1.4.4  # Apache 2.0
 celery==3.1.25  # BSD 3 Clause
 colorlog==2.7.0  # MIT
+coreapi==2.3.1  # BSD
```
laurenbarker: What benefit does the `coreapi` give here?

aaxelb: Hm, whoops, I installed coreapi to check out DRF's [auto-generated docs](http://www.django-rest-framework.org/topics/documenting-your-api/) (which apparently chokes on some of our more unconventional API code, alas), and could have sworn the markdown rendering stopped working when I uninstalled it. I guess I was confused.

laurenbarker: Oh, auto-generated docs would be nice.

laurenbarker:
Looks good 🥇

PR #647, opened 2017-05-08
[Feature] Celery 4

chrisseto:
Keeping a load of messages in RabbitMQ has gotten a bit messy. This change introduces a task that will check the database for harvests that need to be run. It will be added to the beat schedule to run every 5 minutes or so.

~~A lot of the celery bits are a bit messy right now. They'll be fixed in the upgrade to celery 4.~~
This PR also contains the upgrade to celery 4

```
---------- coverage: platform darwin, python 3.5.2-final-0 -----------
Name                             Stmts   Miss  Cover   Missing
--------------------------------------------------------------
share/harvest/__init__.py            1      0   100%
share/harvest/base.py               96      9    91%   104, 124, 139-140, 158, 219-223
share/harvest/exceptions.py          6      0   100%
share/harvest/ratelimit.py          30     15    50%   19-26, 29-30, 33-36, 39
share/harvest/serialization.py      28      2    93%   15, 46
--------------------------------------------------------------
TOTAL                              161     26    84%
```

PR #669, opened 2017-05-25
[Improvement] Add CONTRIBUTING.md

chrisseto:
Pulled out of https://github.com/CenterForOpenScience/SHARE/pull/647

Concerns raised in that PR:

> For consistency with jira, should this be Improvement instead of Enhancement?

> Oh, +1. Is there a reason for Enhancement over Improvement?

> Also, should it be Bug instead of Fix? Fix describes the PR, but Bug describes the ticket/issue... meh, language.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,71 @@
+# CONTRIBUTING
+
+## Style Guide
+
+In the following templates, `TYPE` may be any of `Fix`, `Feature`, `Task`, or `Enhancement`.
+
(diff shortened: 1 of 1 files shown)


laurenbarker:
I would like to point out the irony of porting over the concerns about using "Enhancement" over "Improvement", not addressing the concerns, and then naming the PR:
> [Enhancement] Add CONTRIBUTING.md

chrisseto:
I made the PR and then added everything else. Fixed.

PR #670, opened 2017-05-25
[SHARE-861][Improvement] Consider <location><url> an identifier in MODS

aaxelb:
Even though it's against the spec 😞 

https://openscience.atlassian.net/browse/SHARE-861


laurenbarker:
Tested locally, looks good 💯

PR #671, opened 2017-05-30
[SHARE-832][feature]Add Symbiota harvester

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-832

## Problem
Adding a harvester for Symbiota, harvesting Collections.

## Solution:
- Update setup
- Add the harvester  (Scraping HTML from each record page) 
- Add the transformer 
- Add new source
- Add tests for both harvester/transformer
- Add requests-mock package to requirements


share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+        record_list = []
+        for record in records:
+            record_content = re.findall('collid=(\d+)', record.get('href'))
+            if record_content and record_content[0] not in record_list:
+                record_list.append(record_content[0])
+        total = len(record_list)
+
+        logging.info('Found %d results from swbiodiversity', total)
+        count = 0
+        while count < total:
```
chrisseto: Use a for loop and `enumerate` here.

```python
for count, record in record_list:
    ...
```

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+
+            logger.info('On collection %d of %d (%d%%)', count, total, (count / total) * 100)
+            data = {}
+            identifier = record_list[count]
+
+            data['identifier'] = identifier
+            html = self.requests.get(self.kwargs['list_url'] + '?collid=' + identifier)
```
chrisseto: Use the `furl` library to build URLs. You can look at any other harvester for examples 👍

chrisseto: Use a more descriptive name here like resp or response. html is misleading.

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+                        data['access-rights'] = access_rights
+
+                collection_statistics = start.find_all_next('li')
+                collection_statistics = list(map(self.process_text, collection_statistics))
+                data['collection-statistics'] = self.process_collection_stat(collection_statistics)
+
+            count += 1
+            yield identifier, data
+
+    def process_text(self, text):
```
chrisseto: This method is unnecessary, you can just call `.text()`.
[Don't parse HTML with regex](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags/1732454#1732454)

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+        return re.sub('<[^>]*>', '', str(text)).strip()
+
+    def process_collection_stat(self, list):
+        stat = {}
+        for item in list:
```
chrisseto: Avoid using `list` as a variable name, it shadows the built in type `list`
[Here's a list of builtins/names to avoid](https://docs.python.org/3/library/functions.html).

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+            identifier = record_list[count]
+
+            data['identifier'] = identifier
+            html = self.requests.get(self.kwargs['list_url'] + '?collid=' + identifier)
+            html.raise_for_status()
+            soup = BeautifulSoup(html.content, 'lxml')
+
+            # Peel out script tags and css things to minimize size of HTML
+            for el in itertools.chain(
+                    soup('img'),
```
chrisseto: Looks like this might be double indented?

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,127 @@
...
+                    soup('link', rel=('stylesheet', 'dns-prefetch')),
+                    soup('link', {'type': re.compile('.')}),
+                    soup('noscript'),
+                    soup('script'),
+                    soup(string=lambda x: isinstance(x, Comment)),
+            ):
+                el.extract()
+
+            record = soup.find(id='innertext')
+            title = self.process_text(record.h1)
```
chrisseto: Most, if not all, of this processing should happen in the transformer. You'll want to preserve as much of the original data as possible. If a bug pops up in the parsing, we can just reprocess the data we already have.

chrisseto: If you still prefer to give this data to the transformer, you can massage the data before-hand by overriding `transform`

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,63 @@
+import re
+import logging
+
+from bs4 import BeautifulSoup, Comment
+from furl import furl
+import itertools
```
chrisseto: `itertools` is part of the stdlib

share/transformers/org_swbiodiversity.py
```diff
@@ -0,0 +1,136 @@
+from share.transform.chain import ctx
+from share.transform.chain import links as tools
+from share.transform.chain.parsers import Parser
+from share.transform.chain.soup import SoupXMLTransformer
+from bs4 import BeautifulSoup
+import re
```
chrisseto: These should be ordered as

```python
import re

from bs4 import BeautifulSoup

from share.transform.chain import ctx
from share.transform.chain import links as tools
from share.transform.chain.parsers import Parser
from share.transform.chain.soup import SoupXMLTransformer
```

tests/share/harvesters/test_swbiodiversity_harvester.py
```diff
@@ -0,0 +1,148 @@
+from datetime import timedelta
+
+import requests_mock
```
chrisseto: Could you switch this to use httpretty? @laurenbarker and I thought the interface for it was a bit nicer. It would be good to standardize on a single library.

share/harvesters/org_swbiodiversity.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    """
+    VERSION = 1
+
+    def _do_fetch(self, start, end, **kwargs):
+        end_date = end.date()
+        start_date = start.date()
+        logger.info('Harvesting swbiodiversity %s - %s', start_date, end_date)
+        return self.fetch_records()
```
chrisseto: This appears to disregard date ranges. Did we discuss this at some point?

chrisseto:
@sheriefvt can you fix the merge conflicts?

PR #675, opened 2017-06-02
[SHARE-876][Feature] Allow source creation via API

laurenbarker:
## Questions for reviewer(s)
1. Should update be supported for icons only? The way the system currently works, updating the `long_title` would not be desirable. Is allowing updates to just the icon worth implementing?
2. Is using `long_title` submission to create the other username/label/name the best solution? 

## Purpose 
Allow creation of a source with the API.
See, https://openscience.atlassian.net/browse/SHARE-876

## Changes
* Add `create()` method to `sources/` endpoint which creates ShareUser, Source, and SourceConfig.
* Add `is_trusted` to `ShareUserSerializer` so that trusted users can be created.
* The OSF user will need to have the `add_source` permission.

### Request Content
```json
{
    "data": {
        "type": "Source",
        "attributes": {
            "homePage": "http://super.test.edu",  (optional)
            "longTitle": "Super Test",
            "icon": "<url_for_icon>"
        }
    }
}
```

### Response
```json
{
    "data": {
        "attributes": {
            "home_page": "http://super.test.edu",
            "name": "super_test",
            "long_title": "Super Test"
        },
        "id": "XXXXX-XXX-XXX",
        "relationships": {
            "share_user": {
                "data": {
                    "attributes": {
                        "username": "super_test",
                        "authorization_token": "N2RDMXU397HFIAA19CVZIOM1Q50BUJZQ85WCAFU49WIR46TIYQJ9VZKSLT6AKBXP"
                    },
                    "id": "XXXXX-XXX-XXX",
                    "type": "ShareUser"
                }
            },
            "source_config": {
                "data": {
                    "attributes": {
                        "label": "super_test"
                    },
                    "id": "XXXXX-XXX-XXX",
                    "type": "SourceConfig"
                }
            }
        },
        "type": "Sources"
    }
}
```

### Errors
```json
{
    "errors": [
        {
            "detail": "Could not download/process image. [\"Invalid type. Expected one of ('image/png', 'image/jpeg'). Received text/html\"]",
            "status": "400",
            "source": {
                "pointer": "/data"
            }
        }
    ]
}
```

### DIFF
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -9,6 +9,7 @@ from rest_framework import filters
 from rest_framework.exceptions import ParseError
 from rest_framework.renderers import JSONRenderer
 from rest_framework.parsers import JSONParser
+from rest_framework.serializers import ValidationError
 
 from api import schemas
 from api.pagination import CursorPagination
@@ -40,9 +41,74 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     ordering = ('id', )
     ordering_fields = ('long_title', )
     serializer_class = SourceSerializer
+    permission_classes = [ReadOnlyOrTokenHasScopeOrIsAuthenticated, ]
 
     def get_queryset(self):
         return Source.objects.exclude(icon='').exclude(is_deleted=True)
+
+    def create(self, request, *args, **kwargs):
+        try:
+            home_page = request.data['home_page']
+            long_title = request.data['long_title']
+            # icon = request.data['icon']
+        except KeyError as e:
+            raise ValidationError('{} is a required attribute.'.format(e))
+
+        label = '.'.join(reversed(home_page.split('//')[1].split('.')))
+
+        user_serializer = ShareUserSerializer(
+            data={'username': label, 'is_trusted': True},
+            context={'request': request}
+        )
+
+        if user_serializer.is_valid(raise_exception=True):
+
+            user_instance = user_serializer.save()
+
+            source_instance = Source(
+                user_id=user_instance.id,
+                long_title=long_title,
+                home_page=home_page,
+                name=label
+            )  # icon=icon
+            source_instance.save()
+
+            source_config_instance = SourceConfig(source_id=source_instance.id, label=label)
+            source_config_instance.save()
+
+            return Response(
+                {
+                    'id': source_instance.id,
+                    'type': 'Source',
+                    'attributes': {
+                        'long_title': source_instance.long_title,
+                        'name': source_instance.name,
+                        'home_page': source_instance.home_page
+                    },
+                    'relationships': {
+                        'share_user': {
+                            'data': {
+                                'id': user_instance.id,
+                                'type': 'ShareUser',
+                                'attributes': {
+                                    'username': user_instance.username,
+                                    'authorization_token': user_instance.authorization()
+                                }
+                            }
+                        },
+                        'source_config': {
+                            'data': {
+                                'id': source_config_instance.id,
+                                'type': 'SourceConfig',
+                                'attributes': {
+                                    'label': source_config_instance.label
+                                }
+                            }
+                        }
+                    }
+                },
+                status=status.HTTP_202_ACCEPTED
+            )
 
 
 class NormalizedDataViewSet(viewsets.ModelViewSet):
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -1,5 +1,4 @@
 import jsonschema
-from io import BytesIO
 import requests
 
 from django.db import transaction
@@ -49,12 +48,12 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
 
     queryset = Source.objects.none()  # Required for DjangoModelPermissions
 
+    VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
+
     def get_queryset(self):
         return Source.objects.exclude(icon='').exclude(is_deleted=True)
 
     def create(self, request, *args, **kwargs):
-        VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
-
         try:
             long_title = request.data['long_title']
             icon = request.data['icon']
@@ -62,13 +61,12 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
             raise ValidationError('{} is a required attribute.'.format(e))
 
         try:
-            r = requests.get(icon)
+            r = requests.get(icon, timeout=5)
             header_type = r.headers['content-type'].split(';')[0].lower()
-            if header_type not in VALID_IMAGE_TYPES:
-                raise ValidationError('Invalid type. Expected one of {}. Received {}'.format(VALID_IMAGE_TYPES, header_type))
+            if header_type not in self.VALID_IMAGE_TYPES:
+                raise ValidationError('Invalid type. Expected one of {}. Received {}'.format(self.VALID_IMAGE_TYPES, header_type))
 
-            icon_io = BytesIO(r.content)
-            icon_file = ContentFile(icon_io.getvalue())
+            icon_file = ContentFile(r.content)
         except Exception as e:
             raise ValidationError('Could not download/process image. {}'.format(e))
 
@@ -79,57 +77,51 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
             context={'request': request}
         )
 
-        if user_serializer.is_valid(raise_exception=True):
-            user_instance = user_serializer.save()
+        user_serializer.is_valid(raise_exception=True)
 
-            source_instance = Source(
-                user_id=user_instance.id,
-                long_title=long_title,
-                home_page=request.data['home_page'] if 'home_page' in request.data else None,
-                name=label,
-            )
-            source_instance.icon.save(
-                '{}.tmp'.format(label),
-                content=icon_file,
-                save=True
-            )
+        user_instance = user_serializer.save()
+        source_instance = Source(
+            user_id=user_instance.id,
+            long_title=long_title,
+            home_page=request.data['home_page'] if 'home_page' in request.data else None,
+            name=label,
+        )
+        source_instance.icon.save(label, content=icon_file)
+        source_config_instance = SourceConfig.objects.create(source_id=source_instance.id, label=label)
 
-            source_config_instance = SourceConfig(source_id=source_instance.id, label=label)
-            source_config_instance.save()
-
-            return Response(
-                {
-                    'id': source_instance.id,
-                    'type': 'Source',
-                    'attributes': {
-                        'long_title': source_instance.long_title,
-                        'name': source_instance.name,
-                        'home_page': source_instance.home_page
-                    },
-                    'relationships': {
-                        'share_user': {
-                            'data': {
-                                'id': user_instance.id,
-                                'type': 'ShareUser',
-                                'attributes': {
-                                    'username': user_instance.username,
-                                    'authorization_token': user_instance.accesstoken_set.first().token
-                                }
+        return Response(
+            {
+                'id': source_instance.id,
+                'type': 'Sources',
+                'attributes': {
+                    'long_title': source_instance.long_title,
+                    'name': source_instance.name,
+                    'home_page': source_instance.home_page
+                },
+                'relationships': {
+                    'share_user': {
+                        'data': {
+                            'id': user_instance.id,
+                            'type': 'ShareUser',
+                            'attributes': {
+                                'username': user_instance.username,
+                                'authorization_token': user_instance.accesstoken_set.first().token
                             }
-                        },
-                        'source_config': {
-                            'data': {
-                                'id': source_config_instance.id,
-                                'type': 'SourceConfig',
-                                'attributes': {
-                                    'label': source_config_instance.label
-                                }
+                        }
+                    },
+                    'source_config': {
+                        'data': {
+                            'id': source_config_instance.id,
+                            'type': 'SourceConfig',
+                            'attributes': {
+                                'label': source_config_instance.label
                             }
                         }
                     }
-                },
-                status=status.HTTP_202_ACCEPTED
-            )
+                }
+            },
+            status=status.HTTP_201_CREATED
+        )
 
 
 class NormalizedDataViewSet(viewsets.ModelViewSet):
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -79,20 +79,21 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
 
         user_serializer.is_valid(raise_exception=True)
 
-        user_instance = user_serializer.save()
-        source_instance = Source(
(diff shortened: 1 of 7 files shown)


api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
 
+    def create(self, request, *args, **kwargs):
+        try:
+            home_page = request.data['home_page']
+            long_title = request.data['long_title']
+            # icon = request.data['icon']
+        except KeyError as e:
+            raise ValidationError('{} is a required attribute.'.format(e))
+
+        label = '.'.join(reversed(home_page.split('//')[1].split('.')))
```
chrisseto: Do you think we should continue psuedo-namespacing these for users? IE `providers.com.whatever` or `sources.net.otherthing`

api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+        user_serializer = ShareUserSerializer(
+            data={'username': label, 'is_trusted': True},
+            context={'request': request}
+        )
+
+        if user_serializer.is_valid(raise_exception=True):
```
chrisseto: Could this short-circuit so the rest of this method doesn't have to be indented?

api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+            user_instance = user_serializer.save()
+
+            source_instance = Source(
```
chrisseto: Can you confirm that all requests are in a transaction or place this into one if they are not

api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+            )  # icon=icon
+            source_instance.save()
+
+            source_config_instance = SourceConfig(source_id=source_instance.id, label=label)
+            source_config_instance.save()
+
+            return Response(
+                {
+                    'id': source_instance.id,
+                    'type': 'Source',
```
chrisseto: `Sources`

api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+                long_title=long_title,
+                home_page=home_page,
+                name=label
+            )  # icon=icon
+            source_instance.save()
+
+            source_config_instance = SourceConfig(source_id=source_instance.id, label=label)
+            source_config_instance.save()
+
+            return Response(
```
chrisseto: Can this be built by the serializer?

api/views/workflow.py
```diff
@@ -40,10 +41,75 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+                                'id': source_config_instance.id,
+                                'type': 'SourceConfig',
+                                'attributes': {
+                                    'label': source_config_instance.label
+                                }
+                            }
+                        }
+                    }
+                },
+                status=status.HTTP_202_ACCEPTED
```
chrisseto: `202` is for a request that hasn't been completely processed or has created a job of some sorts. 
`201 CREATED` would be the appropriate response code here

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     ordering = ('id', )
     ordering_fields = ('long_title', )
     serializer_class = SourceSerializer
+    permission_classes = [DjangoModelPermissionsOrAnonReadOnly, ]
+
+    queryset = Source.objects.none()  # Required for DjangoModelPermissions
```
chrisseto: Shouldn't having `get_queryset` defined be the same?

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
 
     def get_queryset(self):
         return Source.objects.exclude(icon='').exclude(is_deleted=True)
 
+    def create(self, request, *args, **kwargs):
+        VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
```
chrisseto: These should be class level.

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+        try:
+            long_title = request.data['long_title']
+            icon = request.data['icon']
+        except KeyError as e:
+            raise ValidationError('{} is a required attribute.'.format(e))
+
+        try:
+            r = requests.get(icon)
```
chrisseto: Can you add a fairly low time out here?

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+            header_type = r.headers['content-type'].split(';')[0].lower()
+            if header_type not in VALID_IMAGE_TYPES:
+                raise ValidationError('Invalid type. Expected one of {}. Received {}'.format(VALID_IMAGE_TYPES, header_type))
+
+            icon_io = BytesIO(r.content)
```
chrisseto: What's the purpose of placing this into a `BytesIO`?

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,141 @@
...
+
+from share.models import Source, ShareUser
+
+
+PROPER_ICON_URL = 'https://staging-cdn.osf.io/preprints-assets/bitss/square_color_no_transparent.png'
+IMPROPER_ICON_URL = 'https://github.com/CenterForOpenScience/SHARE.git'
+INVALID_URL = 'invalidURL'
+
+
+def get_test_data(icon=PROPER_ICON_URL, home_page=None):
```
chrisseto: This should be a `pytest.fixture`, ideally defined in `TestSourcesGet`

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,141 @@
+import json
+import pytest
+
+from share.models import Source, ShareUser
+
+
+PROPER_ICON_URL = 'https://staging-cdn.osf.io/preprints-assets/bitss/square_color_no_transparent.png'
```
chrisseto: We shouldn't be making HTTP requests in tests, can you mock the urls?
There are a couple libraries that are pretty nice, httpretty is my favorite. Sherif has been using mock-requests.

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,141 @@
...
+IMPROPER_ICON_URL = 'https://github.com/CenterForOpenScience/SHARE.git'
+INVALID_URL = 'invalidURL'
```
chrisseto: Can you add a test case for the icon request timing out?

tests/conftest.py
```diff
@@ -68,6 +71,17 @@ def trusted_user():
 
 
 @pytest.fixture
+def source_add_user():
```
chrisseto: This seems specific enough that it could go into test_sources file.

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,141 @@
...
+        test_data = get_test_data(icon=IMPROPER_ICON_URL)
+        resp = client.post(
+            self.endpoint,
+            json.dumps(test_data),
+            content_type='application/vnd.api+json',
+            HTTP_AUTHORIZATION=source_add_user.authorization(),
+        )
+
+        assert resp.status_code == 400
+        assert resp.data[0]['detail'] == 'Could not download/process image. ["Invalid type. Expected one of (\'image/png\', \'image/jpeg\'). Received text/html"]'
```
chrisseto: nitpick: title should be `Could not download/process image` and detail should be `Invalid type....`

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+            user_instance = user_serializer.save()
+
+            source_instance = Source(
+                user_id=user_instance.id,
+                long_title=long_title,
+                home_page=request.data['home_page'] if 'home_page' in request.data else None,
+                name=label,
+            )
+            source_instance.icon.save(
+                '{}.tmp'.format(label),
```
chrisseto: Why does this have the extension `{}.tmp`?

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+                content=icon_file,
+                save=True
```
chrisseto: This doesn't make sense.

api/views/workflow.py
```diff
@@ -40,10 +45,92 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+            )
+
+            source_config_instance = SourceConfig(source_id=source_instance.id, label=label)
```
chrisseto: `SourceConfig.objects.create` might be a bit cleaner

api/views/workflow.py
laurenbarker: One would think. I didn't actually try it without though; just trusted the [docs](http://www.django-rest-framework.org/api-guide/permissions/#djangomodelpermissions):
> If you're using this permission with a view that uses an overridden get_queryset() method there may not be a queryset attribute on the view. In this case we suggest also marking the view with a sentinel queryset, so that this class can determine the required permissions. For example:

api/views/workflow.py
laurenbarker: Maybe partially? It would be three different serializers (one that doesn't exist right now) and the relations aren't usually represented like this. I'll look into it a bit though.

api/views/workflow.py
laurenbarker: It seems like the inflection of all of our types is singular. Do you want that to change? According to the [DRF JSON API docs](http://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#setting-the-resource-name) plural is suppose to be the default so I'm no sure what is happening...

laurenbarker: Could add `JSON_API_PLURALIZE_TYPES = True` to `settings.py`

tests/api/test_sources_endpoint.py
laurenbarker: Turns out adding "title" to the error message would require a custom exception handler (re-writing or patching the JSON API one) and custom exception.

tests/api/test_sources_endpoint.py
laurenbarker: Ended up using HTTPretty even though they don't officially support python3 because mock-requests -- to my knowledge -- doesn't provide a way to mock requests that are in the codebase rather than the test file.

api/views/workflow.py
laurenbarker: Probably not since they don't have a website but I don't really have a preference.

api/views/workflow.py
laurenbarker: It's possible to render a top level with attributes and a related with id and type using [included](http://django-rest-framework-json-api.readthedocs.io/en/stable/usage.html#included) in the serializer. I couldn't get attributes from multiple models to be serialized. Using include on Source would look like:
```json
{
    "data": {
        "type": "Source",
        "id": "1B50E1-C0A-161",
        "attributes": {
            "name": "test_3",
            "homePage": null,
            "longTitle": "Test 3",
            "icon": "/icons/test_3.ico"
        },
        "relationships": {
            "user": {
                "data": {
                    "type": "ShareUser",
                    "id": "1B0D4-553-851"
                }
            }
        }
    }
}
```

api/views/workflow.py
laurenbarker: Pluralizing types breaks 19 API tests so if we switch those would need to be updated.

tests/api/test_sources_endpoint.py
laurenbarker: I'm not sure how to make this a fixture without making a separate fixture for each url type which seems worse.

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,198 @@
...
+
+@pytest.mark.django_db
+class TestSourcesPost:
+    endpoint = '/api/v2/sources/'
+
+    def exceptionCallback(request, uri, headers):
+        time.sleep(6)
+        return (400, headers, uri)
+
+    httpretty.enable()
```
chrisseto: This/these should be encapsulated in a fixture.

tests/api/test_sources_endpoint.py
```diff
@@ -0,0 +1,198 @@
...
+    httpretty.HTTPretty.allow_net_connect = False
+
+    # smallest valid png, from https://github.com/mathiasbynens/small/blob/master/png-transparent.png
```
chrisseto: 🆒

api/views/workflow.py
```diff
@@ -40,10 +44,86 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+                long_title=long_title,
+                home_page=request.data['home_page'] if 'home_page' in request.data else None,
+                name=label,
+            )
+            source_instance.icon.save(label, content=icon_file)
+            source_config_instance = SourceConfig.objects.create(source_id=source_instance.id, label=label)
+
+        return Response(
+            {
+                'id': source_instance.id,
```
chrisseto: I think this should be the encoded ID?

api/views/workflow.py
```diff
@@ -40,10 +44,86 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+
+        try:
+            r = requests.get(icon, timeout=5)
+            header_type = r.headers['content-type'].split(';')[0].lower()
+            if header_type not in self.VALID_IMAGE_TYPES:
+                raise ValidationError('Invalid type. Expected one of {}. Received {}'.format(self.VALID_IMAGE_TYPES, header_type))
+
+            icon_file = ContentFile(r.content)
+        except Exception as e:
+            raise ValidationError('Could not download/process image. {}'.format(e))
```
chrisseto: Can you throw a `logger.exception` in here?

I don't think we should send back the exception here. It could technically be possible to leak data if we receive a chatty exception.

api/views/workflow.py
```diff
@@ -40,10 +44,86 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+            long_title = request.data['long_title']
+            icon = request.data['icon']
+        except KeyError as e:
+            raise ValidationError('{} is a required attribute.'.format(e))
+
+        try:
+            r = requests.get(icon, timeout=5)
+            header_type = r.headers['content-type'].split(';')[0].lower()
+            if header_type not in self.VALID_IMAGE_TYPES:
+                raise ValidationError('Invalid type. Expected one of {}. Received {}'.format(self.VALID_IMAGE_TYPES, header_type))
```
chrisseto: This will be caught by your `except` block, this exception should be raised outside of it.

PR #677, opened 2017-06-07
[SHARE-503][Feature]Add AEA harvester

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-503

## Problem
Add AEA Harvester which is needed to support Registries Phase 1

## Solution:
* Add AEA harvester/transformer
* Add new SHARE source
* Update setup.py
* Add new tests for the harvester and transformer


share/harvesters/org_socialscienceregistry.py
```diff
@@ -0,0 +1,59 @@
...
+
+    def fetch_records(self, start_date, end_date):
+
+        csv_download = self.requests.get(self.kwargs['csv_url'])
+
+        csv_download.raise_for_status()
+
+        decoded_content = csv_download.content.decode('utf-8')
+
+        cr = csv.reader(decoded_content.splitlines(), delimiter=',')
```
chrisseto: You might want to consider using a [DictReader](https://docs.python.org/3/library/csv.html#csv.DictReader) and just sending back each dictionary.

share/harvesters/org_socialscienceregistry.py
```diff
@@ -0,0 +1,59 @@
...
+        logger.info('Harvesting the social science registry %s - %s', start_date, end_date)
+        return self.fetch_records(start_date, end_date)
+
+    def fetch_records(self, start_date, end_date):
+
+        csv_download = self.requests.get(self.kwargs['csv_url'])
+
+        csv_download.raise_for_status()
+
+        decoded_content = csv_download.content.decode('utf-8')
```
chrisseto: Can you verify that the data being received is utf-8 encoded?

share/harvesters/org_socialscienceregistry.py
```diff
@@ -0,0 +1,59 @@
...
+
+        logging.info('Found total %d results from the social science registry', total_records)
+        standard_size = 41
+        records_ignored = 0
+        records_harvested = 0
+
+        for i, record in enumerate(record_list):
+            logger.info('On trial %d of %d (%d%%)', i, total_records, (i / total_records) * 100)
+
+            if (len(record) != standard_size):
```
chrisseto: You don't need the extra parens here.

tests/share/transformers/test_socialscienceregistry_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+        "",
+        ""
+    ]
+}
+'''
+
+
+@pytest.mark.django_db
+def test_func():
+    config = SourceConfig.objects.get(label=('org.socialscienceregistry'))
```
chrisseto: You don't need the parens here.

tests/share/transformers/test_socialscienceregistry_transformer.py
```diff
@@ -0,0 +1,63 @@
...
+        "",
+        "",
+        ""
+    ]
+}
+'''
+
+
+@pytest.mark.django_db
+def test_func():
```
chrisseto: Please use a more specific name.

tests/share/harvesters/test_socialscienceregistry_harvester.py
```diff
@@ -0,0 +1,28 @@
...
+
+csv_repsonse = '''
+...first row is always ignored it contains the columns' titles
+sample2,url2,"June 05, 2017",2017-05-02 16:17:17 -0400,2013-05-21,AEARCTR-0000005,"David James, david@gmail.com",completed,2013-01-26,2014-05-31,"[""electoral"", """"]","","abstract2",2013-03-02,2013-03-07,"information2",See pre-analysis plan.,"Treatment was randomly assigned at the group level, in a nationwide sample.","",Randomization done using Stata.,Individual,N/A,2500 individuals/1600 individuals,N/A,"MDE=5% change in perceived leakage, sd=28.8667, power=0.80, alpha=0.05",Public,This section is unavailable to the public.,,,"",,"","","",info,,"",,"","",
+'''
+
+
+@pytest.mark.django_db
+def test_func():
+    config = SourceConfig.objects.get(label=('org.socialscienceregistry'))
```
chrisseto: Remove extra parens.

share/transformers/org_socialscienceregistry.py
```diff
@@ -0,0 +1,226 @@
...
+class ThroughTags(Parser):
+    tag = tools.Delegate(Tag, ctx)
+
+
+class ThroughSubjects(Parser):
+    subject = tools.Delegate(Subject, ctx)
+
+
+def process_keywords(text):
+    text = re.sub('["]', '', text).strip("[]")[:-2].split(",")
```
chrisseto: Looks like `json.loads` might be a better choice here

share/transformers/org_socialscienceregistry.py
```diff
@@ -0,0 +1,226 @@
+from share.transform.chain import ctx, ChainTransformer
```
chrisseto: Reorder these imports

share/transformers/org_socialscienceregistry.py
```diff
@@ -0,0 +1,226 @@
...
+class Registration(Parser):
+    title = tools.Try(ctx['general-information']['title'])
+    description = tools.Try(ctx['additional-trial-info']['abstract'])
+    date_updated = tools.ParseDate(tools.Try(ctx['general-information']['last-updated']))
+    date_published = tools.ParseDate(tools.Try(ctx['general-information']['published-at']))
+    related_agents = tools.Map(tools.Delegate(Creator), tools.Try(ctx.pi))
+    identifiers = tools.Map(
+        tools.Delegate(WorkIdentifier),
+        tools.Concat(
+            tools.Try(ctx['general-information']['url'])
```
laurenbarker: The IRI parser should probably be moved down here so that a through identifier isn't created if the given identifier isn't valid.

share/sources/org.socialscienceregistry/source.yaml
```diff
@@ -0,0 +1,16 @@
...
+  harvester: org.socialscienceregistry
+  harvester_kwargs:
+      csv_url: https://www.socialscienceregistry.org/trials/search.csv?query=
+  label: org.socialscienceregistry
+  rate_limit_allowance: 1
+  rate_limit_period: 2
+  transformer: org.socialscienceregistry
+  transformer_kwargs: {}
+home_page: https://www.socialscienceregistry.org
+long_title: AEA RCT Registry
```
laurenbarker: Have we already checked with product on what the long title should be for AEA?

share/harvesters/org_socialscienceregistry.py
```diff
@@ -0,0 +1,58 @@
...
+
+        logging.info('Found total %d results from the social science registry', total_records)
+        standard_size = 41
+        records_ignored = 0
+        records_harvested = 0
+
+        for i, record in enumerate(record_list):
+            logger.info('On trial %d of %d (%d%%)', i, total_records, (i / total_records) * 100)
+
+            if len(record) != standard_size:
```
laurenbarker: @sheriefvt, why are records that aren't of length 41 incorrect?

sheriefvt: @laurenbarker, unfortunately, some of the records have new lines and/or extra/lesser number of commas. Causing reader to split the record into multiple records with sizes less than the actual which should be 41

laurenbarker:
Ran locally and didn't see any errors 🥇  There where more incorrect than correct records though so our number probably won't get near the reported number on their site.

share/harvesters/org_socialscienceregistry.py
aaxelb: Are there invalid records not caused by newlines inside quotes?

sheriefvt: @laurenbarker @aaxelb the total records ignored are calculated over the entire record set of AEA not the ones in the date range only. So if a record with the last update date is out of range and was incorrectly CSV formatted too, then it will count as ignored.

chrisseto: @sheriefvt it looks like python will handle the new lines just find if you don't do anything:
```python
ipdb> data = list(csv.reader(csv_response.text.split('\n')))
ipdb> [x for x in data if len(x) != 41]
[[]]
```

chrisseto:
@sheriefvt, this somehow got lost amongst the sprints. I added it back in to the current sprint. Could you resolve the conflicts?

sheriefvt:
@chrisseto done 👍

PR #684, opened 2017-06-16
[SHARE-686][Feature] IngestJob and Regulator

aaxelb:
In broad strokes:
- Jobs
  - Rename `HarvestLog` to `HarvestJob` (lots of this PR is just replacing `log` with `job`)
  - Add `IngestJob`, which keeps track of the ingestion status of a given raw datum
  - Make `harvest` task logic reusable by refactoring into `JobConsumer`, with children `HarvestJobConsumer`, `IngestJobConsumer`
  - Replace `transform` task with `ingest`, which grabs an available (or given) IngestJob, then transforms, regulates, and (in a future PR) saves the result as a single-source graph
    - For now, preserves legacy behavior by saving the regulated graph as a NormalizedData and spawning a `disambiguate` task for it
- Regulator
  - Add a basic framework for regulation, allowing new regulation steps that operate on a MutableGraph from a transformer
  - Right now, just one validation step which runs `JSONLDValidator`
- `sharectl ingest`
  - Create IngestJobs for specific raw data by ID, or for all data from one or more source configs.


setup.py
```diff
@@ -93,6 +93,13 @@
             'org.neurovault = share.harvesters.org_neurovault:NeuroVaultHarvester',
             'org.plos = share.harvesters.org_plos:PLOSHarvester',
             'org.swbiodiversity = share.harvesters.org_swbiodiversity:SWHarvester',
-        ]
+        ],
+        'share.regulate.node_steps': [
```
chrisseto: `share.regulate.steps.node`
`share.regulate.steps.graph`
`share.regulate.steps.validation`
?

share/bin/harvest.py
```diff
@@ -107,18 +107,18 @@ def harvest(args, argv):
         print(datum)
 
 
-@command('Create harvestlogs for the specified SourceConfig')
+@command('Create harvestjobs for the specified SourceConfig')
```
chrisseto: `HarvestJobs`

share/bin/ingest.py
```diff
@@ -33,3 +34,49 @@ def transform(args, argv):
             print('Parsed raw data "{}" into'.format(name))
             pprint(transformer.transform(data))
             print('\n')
+
+
+@command('Create IngestJobs for the specified raw data/um')
```
chrisseto: `s/raw data\/um/RawDatum(s)/`?

share/bin/ingest.py
```diff
@@ -33,3 +34,49 @@ def transform(args, argv):
...
+def ingest(args, argv):
+    """
+    Usage: {0} ingest (<source_configs>... | --all) [--superfluous] [--task | --run]
+           {0} ingest --ids <raw_data_ids>... [--task | --run]
+
+    Options:
+        -i, --ids           Provide RawDatum IDs to ingest specifically
+        -s, --superfluous   Reingest datums that already have an IngestJob
```
chrisseto: RawDatums or data?

share/harvest/base.py
```diff
@@ -115,7 +114,7 @@ def fetch_date(self, date: datetime.date, **kwargs):
         """
         return self.fetch_date_range(date - datetime.timedelta(days=1), date, **kwargs)
 
-    def fetch_date_range(self, start, end, limit=None, **kwargs):
+    def fetch_date_range(self, start, end, limit=None):
```
chrisseto: The kwargs are there to allow passing arbitrary keyword arguments to `_do_fetch` without having to override any other functions. 
If additional kwargs are passed in, they should eventually get rejected by `_do_fetch`.
Is there any reason to remove them?

share/harvest/base.py
```diff
@@ -147,15 +146,17 @@ def fetch_date_range(self, start, end, limit=None, **kwargs):
...
 
         if not isinstance(data_gen, types.GeneratorType) and len(data_gen) != 0:
             raise TypeError('{!r}._do_fetch must return a GeneratorType for optimal performance and memory usage'.format(self))
 
         for i, blob in enumerate(data_gen):
             result = FetchResult(blob[0], self.serializer.serialize(blob[1]), *blob[2:])
 
-            if result.datestamp and (result.datestamp.date() < start.date() or result.datestamp.date() > end.date()):
+            if result.datestamp is None:
+                result.datestamp = end
```
chrisseto: This feels like shakey ground, why not leave it blank?

share/models/ingest.py
```diff
@@ -132,10 +132,18 @@ def natural_key(self):
         return (self.label,)
 
     def get_harvester(self, **kwargs):
-        return self.harvester.get_class()(self, **{**kwargs, **(self.harvester_kwargs or {})})
+        """Return a harvester instance configured for this SourceConfig.
+
+        Passed kwargs override or add to harvester_kwargs.
+        """
+        return self.harvester.get_class()(self, **{**(self.harvester_kwargs or {}), **kwargs})
```
chrisseto: I'm not sure how I feel about passing options here rather to the fetch/harvest methods. I think it make sense to have both.
`set_spec` feels like it would be more appropriate as a per call parameter rather than a per instance parameter.

share/models/ingest.py
```diff
@@ -372,10 +380,10 @@ class RawDatum(models.Model):
     # The sha256 of the datum
     sha256 = models.TextField(validators=[validators.MaxLengthValidator(64)])
 
-    datestamp = models.DateTimeField(null=True, help_text=(
+    datestamp = models.DateTimeField(default=pendulum.now, help_text=(
         'The most relevant datetime that can be extracted from this RawDatum. '
-        'This may be, but is not limitted to, a deletion, modification, publication, or creation datestamp. '
-        'Ideally, this datetime should be appropriate for determining the chronological order it\'s data will be applied.'
+        'This may be, but is not limited to, a deletion, modification, publication, or creation datestamp. '
+        'Ideally, this datetime should be appropriate for determining the chronological order its data will be applied. '
```
chrisseto: Trailing space?

share/models/ingest.py
```diff
@@ -372,10 +380,10 @@ class RawDatum(models.Model):
     # The sha256 of the datum
     sha256 = models.TextField(validators=[validators.MaxLengthValidator(64)])
 
-    datestamp = models.DateTimeField(null=True, help_text=(
+    datestamp = models.DateTimeField(default=pendulum.now, help_text=(
```
chrisseto: Should this just be `auto_now=True` or `auto_add_now=True`? It will take into account the system's timezone, or is that being ignored on purpose? `pendulum.now` defaults to UTC, correct?

share/models/jobs.py
```diff
@@ -198,12 +205,12 @@ class SkipReasons(enum.Enum):
     date_created = models.DateTimeField(auto_now_add=True, editable=False)
     date_modified = models.DateTimeField(auto_now=True, editable=False, db_index=True)
 
-    source_config = models.ForeignKey('SourceConfig', editable=False, related_name='harvest_logs', on_delete=models.CASCADE)
+    source_config = models.ForeignKey('SourceConfig', editable=False, related_name='%(class)ss', on_delete=models.CASCADE)
```
chrisseto: :( no way to make this `jobtype_jobs`?

share/regulate/steps/__init__.py
```diff
@@ -0,0 +1,54 @@
+from share.models import RegulatorLog
+from share.regulate.errors import RegulatorError
+
+
+class BaseStep:
```
chrisseto: +9001

setup.py
aaxelb: Yeah, that's better.

share/harvest/base.py
aaxelb: Right now this is only used for `set_spec`, which is already in `harvester_kwargs`, so the OAI harvester needs extra logic to manually handle the override (`set_spec = set_spec or self.set_spec`). I think the BaseHarvester API would be better and easier to work with if harvester-specific kwargs enter in only one place (`__init__`, stored in `self.kwargs` by default), not two.

When using a harvester, it's just a change from
```
harvester = config.get_harvester()
harvester.fetch(set_spec='my-crazy-setspec')
```
to
```
harvester = config.get_harvester(set_spec='my-crazy-setspec')
harvester.fetch()
```
which seems like no sacrifice. It'd also be nice to guarantee that any kwargs you can provide manually can also be specified in `harvester_kwargs`, and vice versa.

share/harvest/base.py
aaxelb: I wanted a non-null datestamp so we could reliably figure out the most recent RawDatum for a suid. If the transformer didn't give a useful datestamp, it seemed like a safe assumption the datum was relevant within the harvested range, rather than using `now` or something. Would `start` be better?

share/models/ingest.py
aaxelb: `auto_now*` will override values set before saving, which is bad. The default here is just a fallback if it somehow got this far without a sensible datestamp.

`pendulum.now` defaults to UTC, yes, but does that matter? 08:00-05 is the same as 13:00+00 is the same as 17:00+04

share/models/ingest.py
aaxelb: I have no idea how that happened.

share/models/jobs.py
aaxelb: If you can figure out a way, please let me know. `harvestjobs` makes me sad.

aaxelb: (a custom foreign key field could probably do it, but that feels gross too)

share/models/ingest.py
aaxelb: I think I found a good compromise -- preserve passing kwargs to fetch/harvest methods, but also pass the config's harvester_kwargs into `_do_fetch` the same way (and let passed kwargs override those in the config). It unifies the kwargs, lets all of them be per-call, and gets rid of `self.kwargs`, which made me feel dirty and full of regret.

PR #685, opened 2017-06-19
[SHARE-925][Feature] Allow case insensitive keyword search

laurenbarker:
## Purpose
Allow case insensitive search for sources.

## Changes
Add a normalizer for keywords.

## Side effects
Need to re-index everything.

### DIFF
--- a/bots/elasticsearch/bot.py
+++ b/bots/elasticsearch/bot.py
@@ -55,6 +55,14 @@ class ElasticSearchBot:
                         'autocomplete_filter'
                     ]
                 },
+            },
+            # Normalizers are not officially supported
+            # https://www.elastic.co/guide/en/elasticsearch/reference/5.4/analysis-normalizers.html
+            'normalizer': {
+                'lowercasenormalizer': {
+                    'type': 'custom',
+                    'filter': ['lowercase']
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


aaxelb:
Does using a normalizer like this mean the sources facet/donut will be all lower case?

laurenbarker:
Ew. Yes it does.

aaxelb:
Maybe we should add a `sources.exact` field for aggregations/filtering? Then once preprints/dashboard/whoever have switched over, `sources` could be `text`, which would be nicer for typed queries.

laurenbarker:
That would probably make the most sense.

laurenbarker:
@chrisseto would you be okay with splitting the case insensitive search stuff into two releases? One for adding an exact field for sources and another for making the sources field a text field. It would probably be easier than coordinating a release with preprints and the dashboard.

chrisseto:
Would it make more sense to use an [analyzer/tokenizer](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-analyzer.html) given that they are officially supported?

laurenbarker:
@chrisseto, analyzer/tokenizer can't be used on a [keyword field](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/keyword.html)
> Keyword fields are only searchable by their exact value.

So are you saying your on board with switching to a text field or am I missing something?

chrisseto:
@laurenbarker Do you remember what the final decision on this was?

laurenbarker:
@chrisseto, we still haven't heard back from product on this one. I would say move it out of the sprint and close the PR until a decision is made.

chrisseto:
@laurenbarker alright, thank you!

PR #686, opened 2017-06-21
[Fix] Use RabbitMQ to index creativeworks on the fly and mitigate rac…

chrisseto:
…e conditions

### DIFF
--- a/share/search/daemon.py
+++ b/share/search/daemon.py
@@ -1,0 +1,127 @@
+import logging
+import time
+from queue import Empty
+
+from django.conf import settings
+
+from share import util
+
+from bots.elasticsearch.tasks import index_model
+
+
+logger = logging.getLogger(__name__)
+
+
+class SearchIndexerDaemon:
+
+    def __init__(self, celery_app, max_size=500, timeout=5, flush_interval=10):
+        self.app = celery_app
+        self.messages = []
+        self.last_flush = 0
+
+        self.flush_interval = flush_interval
+        self.max_size = max_size
+        self.timeout = timeout
+        self._running = False
+
+        self.sentry = None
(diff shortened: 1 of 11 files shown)


share/tasks.py
```diff
@@ -86,20 +87,45 @@ def disambiguate(self, normalized_id):
...
             countdown=(random.random() + 1) * min(settings.CELERY_RETRY_BACKOFF_BASE ** self.request.retries, 60 * 15)
         )
 
+    # Only index creativeworks on the fly, for the moment.
+    data = [x.id for x in updated or [] if isinstance(x, AbstractCreativeWork)]
+
+    if not data:
+        return
+
+    try:
```
aaxelb: This should be moved to a function (or something) in `share.search`, so it's easy to index works (or eventually anything) from anywhere.

share/search/daemon.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def _run(self, queue):
+        self.last_flush = time.time()
+
+        while True:
+            try:
+                message = queue.get(timeout=self.timeout)
+                self.messages.append(message)
+                # TODO Support more than just creativeworks?
+                self.buffer.extend(message.payload.get('CreativeWork', []))
```
aaxelb: Why keep a separate buffer of works, rather than extracting the works in `flush()`? More buffers more problems.

Would there be a downside if `max_size` limited `messages` instead?

chrisseto: Technically speaking there is no limit to the number of works that a single message could contain. Just a precaution 🤷‍♂️

aaxelb: The message itself just contains IDs, though, so long lists don't seem like a problem. I'm mostly worried about the added complexity when we add support for other models. We'd have to maintain a separate buffer for each model, or manage a shared buffer for all models, or something similarly gross.

PR #688, opened 2017-06-23
[SHARE-912][Feature]Add JSON encrypted field to sourceconfig

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-912

## Problem
Add a encrypted JSON field to the SourceConfig model for storing sensitive information, like API keys.
Why: Whenever a harvester is added that requires API keys, we need to edit the deployments to securely distribute those keys. This will allow us to deploy harvesters faster and reduce the amount of time spent coordinating with devops

## Solution:

* Add TestEncryptedJsonField to util
* Add new field json_encrypt of type TestEncryptedJsonField to the sourceconfig model
* Add salt value to share settings
* Add tests to the new field type
* Add Django migrations


share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+from cryptography.exceptions import InvalidTag
+from django.db import models
+from django.conf import settings
+import jwe
+
+SENSITIVE_DATA_KEY = jwe.kdf(settings.SECRET_KEY.encode('utf-8'), settings.SALT.encode('utf-8'))
+
+
+class EncryptedJSONField(models.BinaryField):
+    '''
```
chrisseto: Prefer triple double quotes (`"""`) for doc strings.

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+    This field transparently encrypts data in the database. It should probably only be used with PG unless
+    the user takes into account the db specific trade-offs with TextFields.
+    '''
+    prefix = 'jwe:::'
```
chrisseto: If you declare `prefix` as bytes (`b'jwe:::'`) you won't have to call `.encode(` on it.

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def get_db_prep_value(self, input_json, **kwargs):
+        if input_json:
+            input_json = json.dumps(input_json).encode('utf-8')
+            try:
+                input_json = self.prefix.encode('utf-8') + jwe.encrypt(bytes(input_json), SENSITIVE_DATA_KEY)
+            except InvalidTag:
+                # Allow use of an encrypted DB locally without encrypting fields
+                if settings.DEBUG_MODE:
```
chrisseto: Given that we have default keys we can just remove this block

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+                    raise
+        return input_json
+
+    def to_python(self, output_json):
+        if output_json:
+            try:
+                output_json = jwe.decrypt(bytes(output_json[len(self.prefix.encode('utf-8')):]), SENSITIVE_DATA_KEY)
+            except InvalidTag:
+                # Allow use of an encrypted DB locally without decrypting fields
+                if settings.DEBUG_MODE:
```
chrisseto: Given that we have default keys we can just remove this block

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+                    pass
+                else:
+                    raise
+        return input_json
+
+    def to_python(self, output_json):
+        if output_json:
+            try:
+                output_json = jwe.decrypt(bytes(output_json[len(self.prefix.encode('utf-8')):]), SENSITIVE_DATA_KEY)
```
chrisseto: No need to encode `prefix` here the `len` will be the same.

chrisseto: Rather than casting `output_json` to `bytes` call `.decode('utf-8')` on it

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
...
+    This field transparently encrypts data in the database. It should probably only be used with PG unless
+    the user takes into account the db specific trade-offs with TextFields.
+    '''
+    prefix = 'jwe:::'
+
+    def get_db_prep_value(self, input_json, **kwargs):
+        if input_json:
+            input_json = json.dumps(input_json).encode('utf-8')
+            try:
+                input_json = self.prefix.encode('utf-8') + jwe.encrypt(bytes(input_json), SENSITIVE_DATA_KEY)
```
chrisseto: Rather than casting `input_json` to bytes call .decode('utf-8') on it

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
+import json
+
+from cryptography.exceptions import InvalidTag
```
chrisseto: Can you re-order these imports?

```
import stdlib

import 3rdpartylib1

import 3rdpartylib2

import locallibs
```

share/migrations/0043_merge_20170623_1531.py
```diff
@@ -0,0 +1,16 @@
+# -*- coding: utf-8 -*-
```
chrisseto: You can avoid making merge migrations by deleted your old one and letting django recreate it.

tests/share/util/test_encrypted_field.py
```diff
@@ -0,0 +1,40 @@
+import pytest
+
+from share.util.fields import EncryptedJSONField
+
+
+class TestEncryptedJsonField:
+    @pytest.fixture
```
chrisseto: Newline here, please

share/util/fields.py
```diff
@@ -0,0 +1,45 @@
+import json
```
chrisseto: This should go into `share/models/fields.py`

share/models/fields.py
```diff
@@ -550,3 +555,37 @@ def bulk_related_objects(self):
...
+
+class EncryptedJSONField(models.BinaryField):
+    """
+    This field transparently encrypts data in the database. It should probably only be used with PG unless
+    the user takes into account the db specific trade-offs with TextFields.
+    """
+    prefix = b'jwe:::'
+
+    def get_db_prep_value(self, input_json, **kwargs):
+        if input_json:
```
chrisseto: This is a personal preference, in situations like these it is better practice to short-circuit.
Rather than having this entire method nested in an `if` block, just return early. It makes it easier to follow the logic.

Could you also change `to_python`?

chrisseto: The return value here should be more deterministic. Returning `input_json` could have some weird side effects.

If `input_json` is an empty dict (`{}`) it will attempt to store that python object in the database. The same goes for empty lists and empty strings. Returning `None` would ensure that strange data can't sneak into the database.

share/models/fields.py
```diff
@@ -550,3 +555,37 @@ def bulk_related_objects(self):
...
+            input_json = json.dumps(input_json).encode('utf-8')
+            try:
+                input_json = self.prefix + jwe.encrypt(input_json, SENSITIVE_DATA_KEY)
+
+            except InvalidTag:
```
chrisseto: You can remove this entire try block, if `raise` is the only statement in the except it effectively does nothing.

share/models/fields.py
```diff
@@ -550,3 +555,37 @@ def bulk_related_objects(self):
...
+                raise
+        return input_json
+
+    def to_python(self, output_json):
+        if output_json:
+            try:
+                output_json = jwe.decrypt(bytes(output_json[len(self.prefix):]), SENSITIVE_DATA_KEY)
+            except InvalidTag:
+                # Allow use of an encrypted DB locally without decrypting fields
+                if settings.DEBUG_MODE:
```
chrisseto: You can remove this logic the same way you did in `get_db_prep_value`

share/models/fields.py
```diff
@@ -550,3 +555,37 @@ def bulk_related_objects(self):
...
+                    pass
+                else:
+                    raise
+            output_json = json.loads(output_json.decode("utf-8"))
```
chrisseto: Single quotes here, please.

share/models/fields.py
```diff
@@ -550,3 +555,37 @@ def bulk_related_objects(self):
...
+                input_json = self.prefix + jwe.encrypt(input_json, SENSITIVE_DATA_KEY)
+
+            except InvalidTag:
+                raise
+        return input_json
+
+    def to_python(self, output_json):
+        if output_json:
+            try:
+                output_json = jwe.decrypt(bytes(output_json[len(self.prefix):]), SENSITIVE_DATA_KEY)
```
chrisseto: Use `.decode` here rather than casting to bytes.

tests/share/util/test_encrypted_field.py
```diff
@@ -0,0 +1,41 @@
...
+from share.models.fields import EncryptedJSONField
+
+
+class TestEncryptedJsonField:
+
+    @pytest.fixture
+    def field(self):
+        return EncryptedJSONField(null=True, blank=True)
+
+    def test_encrypt_and_decrypt_list(self, field):
```
chrisseto: Could you add some tests cases for storing `[]`, `{}`, `set()`, `None`, and `''`

You can use pytest's [parametrize](https://docs.pytest.org/en/latest/parametrize.html) to avoid duplicating a lot of this logic.

share/models/fields.py
```diff
@@ -550,3 +554,30 @@ def bulk_related_objects(self):
...
+
+        input_json = self.prefix + jwe.encrypt(json.dumps(input_json).encode('utf-8'), SENSITIVE_DATA_KEY)
+
+        return input_json
+
+    def to_python(self, output_json):
+        if not output_json:
+            return None
+
+        output_json = json.loads(jwe.decrypt(output_json[len(self.prefix):], SENSITIVE_DATA_KEY).decode("utf-8"))
```
chrisseto: Single quotes here, please.

share/models/ingest.py
```diff
@@ -126,6 +126,8 @@ class SourceConfig(models.Model):
 
     disabled = models.BooleanField(default=False)
 
+    json_encrypt = EncryptedJSONField(blank=True, null=True)
```
chrisseto: Could you replace the `transformer_kwargs` and `harvester_kwargs` with the encrypted field?
You'll have to write a migration for it. I think the best way would be:
* Add a new encrypted column/field with different names
* Iterate over all the rows and copy the settings over, in python
* Drop the old columns
* Rename the new ones

share/models/fields.py
```diff
@@ -24,6 +26,8 @@
 
 from db.deletion import DATABASE_CASCADE
 
+SENSITIVE_DATA_KEY = jwe.kdf(settings.SECRET_KEY.encode('utf-8'), settings.SALT.encode('utf-8'))
```
chrisseto: Can you move this into `settings.py`? Generally speaking you'll want to keep constant values in `settings.py`

share/models/ingest.py
sheriefvt: done..thanks for feedback

share/migrations/0045_sourceconfig_json_encrypt.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
chrisseto: Could you remove this migration, it is no longer needed.

chrisseto: You'll have to recreate migration 0046.

laurenbarker:
@chrisseto, was this merged via another PR?

chrisseto:
It was not, it somehow didn't get merged for as part of 2.11 so we bumped it to 2.12

laurenbarker:
@sheriefvt, would you mind resolving the merge conflicts so we can get this merged?

sheriefvt:
@laurenbarker sorry forgot about this..will do today

PR #689, opened 2017-06-23
[RANDOM][Feature] Allow long_title query param on sources endpoint

laurenbarker:
## Purpose
Misc feature that could be useful if we ever do a source detail page.

~~The sources page shows more sources than the source donut on the discover page. Since we can't tell if a source has data in elastic search using the API, the sources page will need to be based on an elastic aggregation. Favicons aren't in the elastic results. To reduce the number of requests, the sources endpoint accepts a query parameter to specify which sources to return. The frontend can do an elastic search aggregation, get those source names, and then make a request for only those sources.~~

## Changes
Allow `long_title` query parameter for the sources endpoint. 

## Side-effects
None

### DIFF
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -6,6 +6,7 @@ import requests
 from django.db import transaction
 from django.utils import timezone
 from django.core.files.base import ContentFile
+from django_filters.rest_framework import DjangoFilterBackend
 
 from rest_framework import viewsets, views, status
 from rest_framework.response import Response
@@ -48,13 +49,12 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     ordering_fields = ('long_title', )
     serializer_class = SourceSerializer
     permission_classes = [DjangoModelPermissionsOrAnonReadOnly, ]
+    filter_backends = (DjangoFilterBackend,)
+    filter_fields = ('long_title',)
 
-    queryset = Source.objects.none()  # Required for DjangoModelPermissions
+    queryset = Source.objects.exclude(icon='').exclude(is_deleted=True)  # Required for DjangoModelPermissions
 
     VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
-
-    def get_queryset(self):
-        return Source.objects.exclude(icon='').exclude(is_deleted=True)
 
     def create(self, request, *args, **kwargs):
         try:
--- a/share/models/ingest.py
+++ b/share/models/ingest.py
@@ -69,6 +69,7 @@ class Source(models.Model):
     name = models.TextField(unique=True)
     long_title = models.TextField(unique=True)
     home_page = models.URLField(null=True)
+    # Sources without an icon are excluded from the API
     icon = models.ImageField(upload_to=icon_name, storage=SourceIconStorage(), blank=True)
     is_deleted = models.BooleanField(default=False)
 
--- a/tests/api/test_sources_endpoint.py
+++ b/tests/api/test_sources_endpoint.py
@@ -3,6 +3,7 @@ import pytest
 import time
 
 import httpretty
+from furl import furl
 
 from django.contrib.auth.models import Permission
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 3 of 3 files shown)


api/views/workflow.py
```diff
@@ -49,12 +50,21 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
     VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
 
     def get_queryset(self):
-        return Source.objects.exclude(icon='').exclude(is_deleted=True)
+        queryset = self.queryset
+        source_names = self.request.query_params.get('long_titles', None)
+
+        if source_names is None:
+            return queryset
+        source_names = source_names.split(',')
```
aaxelb: Is it enforced anywhere that `long_title` doesn't contain a comma?

api/views/workflow.py
```diff
@@ -49,12 +50,21 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
+        source_filter_qs = Q()
+        for name in source_names:
+            source_filter_qs = source_filter_qs | Q(long_title=name)
```
aaxelb: Would `queryset.filter(long_title__in=source_names)` work instead?

api/views/workflow.py
aaxelb: >OpenKnowledge@NAU - Search the scholarly, creative and instructional output of NAU
University of Minnesota, Digital Conservancy
University of Illinois at Urbana-Champaign, IDEALS

chrisseto:
There are hard limits on the total length of a URL in some places (Nginx/Chrome/Etc), we might actually hit that...

api/views/workflow.py
laurenbarker: Oh shoot, didn't think of that

laurenbarker:
So make an elastic request on the backend every time the API is hit?

aaxelb:
I was thinking just filter in the browser... does this endpoint need to be filtered? Or does only the displayed list matter?

laurenbarker:
Oh word, only the display list matters. I still think getting a source by long_title would be useful though so I might just modify this PR a bit so it only accepts one long_title (since the works already done).

api/views/workflow.py
```diff
@@ -49,12 +49,16 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
...
 
     VALID_IMAGE_TYPES = ('image/png', 'image/jpeg')
 
     def get_queryset(self):
-        return Source.objects.exclude(icon='').exclude(is_deleted=True)
+        long_title = self.request.query_params.get('long_title', None)
+
+        if long_title is None:
+            return self.queryset
+        return self.queryset.filter(long_title=long_title)
```
aaxelb: Since it only checks field equality, couldn't we use `DjangoFilterBackend` instead of a custom `get_queryset`?

api/views/workflow.py
```diff
@@ -48,14 +49,13 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     ordering_fields = ('long_title', )
     serializer_class = SourceSerializer
     permission_classes = [DjangoModelPermissionsOrAnonReadOnly, ]
+    filter_backends = (DjangoFilterBackend,)
+    filter_fields = ('long_title',)
 
-    queryset = Source.objects.none()  # Required for DjangoModelPermissions
+    queryset = Source.objects.exclude(icon='').exclude(is_deleted=True)  # Required for DjangoModelPermissions
```
aaxelb: Would you mind adding a comment (or description?) on the `icon` field definition that explains how sources without icons aren't exposed in the api/front end? It's unintuitive, easy to forget, and unwritten besides this exclude.

laurenbarker: Sure thing 📝

share/models/ingest.py
```diff
@@ -69,6 +69,7 @@ class Source(models.Model):
     name = models.TextField(unique=True)
     long_title = models.TextField(unique=True)
     home_page = models.URLField(null=True)
+    # Sources without an icon are excluded from the API
```
laurenbarker: @aaxelb, is this clear enough?

aaxelb: Sure 👍

laurenbarker:
Closing since this PR is no longer relevant. We should probably support basic query parameters at some point.

PR #691, opened 2017-06-26
Hotfix/2.9.7

chrisseto:


share/migrations/0040_rawdatum_no_change.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('share', '0039_auto_20170614_1825'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rawdatum',
+            name='no_change',
+            field=models.NullBooleanField(help_text='Indicates that this RawDatum does not contain any new information. This allows the RawDataJanitor to find records that have not been processed.Records that do not contain changes will no have a NormalizedData associated with them, which would otherwise look like data that has not yet been processed.'),
```
laurenbarker: "will no have a NormalizedData" --> "will not have a NormalizedData"

share/models/ingest.py
```diff
@@ -380,6 +381,13 @@ class RawDatum(models.Model):
     date_modified = models.DateTimeField(auto_now=True, editable=False)
     date_created = models.DateTimeField(auto_now_add=True, editable=False)
 
+    no_change = models.NullBooleanField(null=True, help_text=(
+        'Indicates that this RawDatum does not contain any new information. '
+        'This allows the RawDataJanitor to find records that have not been processed.'
+        'Records that do not contain changes will no have a NormalizedData associated with them, '
```
laurenbarker: Source of the typo: "will no have a NormalizedData" --> "will not have a NormalizedData"

share/janitor/tasks.py
```diff
@@ -14,7 +18,24 @@ def rawdata_janitor(self, limit=100):
...
+    #         Join Filter: (share_sourceuniqueidentifier.source_config_id = share_sourceconfig.id)
+    #         ->  Nested Loop  (cost=1.13..Big Numbers)
+    #               ->  Nested Loop Anti Join  (cost=0.56..Big Numbers)
+    #                     ->  Seq Scan on share_rawdatum  (cost=0.00..Big Numbers)
+    #                     ->  Index Only Scan using share_normalizeddata_c0e72696 on share_normalizeddata  (cost=0.56..Small Numbers)
+    #                           Index Cond: (raw_id = share_rawdatum.id)
+
+    qs = RawDatum.objects.select_related('suid__source_config').annotate(
+        has_normalizedata=Exists(NormalizedData.objects.values('id').filter(raw=OuterRef('id'))),
+    ).exclude(no_change=True, has_normalizedata=True)
```
laurenbarker: Awk.

share/janitor/tasks.py
```diff
@@ -2,19 +2,40 @@
...
+    # Limit  (cost=1.13..Much Smaller Numbers)
+    #   ->  Nested Loop  (cost=1.13..Big Numbers)
+    #         Join Filter: (share_sourceuniqueidentifier.source_config_id = share_sourceconfig.id)
+    #         ->  Nested Loop  (cost=1.13..Big Numbers)
+    #               ->  Nested Loop Anti Join  (cost=0.56..Big Numbers)
+    #                     ->  Seq Scan on share_rawdatum  (cost=0.00..Big Numbers)
+    #                     ->  Index Only Scan using share_normalizeddata_c0e72696 on share_normalizeddata  (cost=0.56..Small Numbers)
+    #                           Index Cond: (raw_id = share_rawdatum.id)
+
+    qs = RawDatum.objects.select_related('suid__source_config').annotate(
```
aaxelb: Since `transform` needs only the id, looks like this `select_related` is only used by `logger.debug`. Seems like you could further speed this up by skipping these joins and getting a `values_list` of ids.

share/janitor/tasks.py
```diff
@@ -2,19 +2,40 @@
...
 
 @celery.shared_task(bind=True)
-def rawdata_janitor(self, limit=100):
+def rawdata_janitor(self, limit=500):
     """Find RawDatum that do not have a NormalizedData process them
     """
     count = 0
-    for rd in RawDatum.objects.select_related('suid__source_config').filter(normalizeddata__isnull=True).order_by('id')[:limit].iterator():
+
+    # NOTE: Do NOT use .iterator here. It will create a temporary table and eat disk space like no other
```
aaxelb: Was `.iterator` the problem, or was it the join from `.filter(normalizeddata__isnull=True)`? I'm trying to better understand nuances of Django and SQL, and Django [claims that](https://docs.djangoproject.com/en/1.11/ref/models/querysets/#iterator)
> internally, the default iterator calls iterator() and caches the return value

but I wouldn't be shocked if something weirder was going on.

share/janitor/tasks.py
```diff
@@ -2,19 +2,40 @@
...
+    # the limit lets this query fit nicely in memory and actually finish executing
+    # Be very careful about changing this query. If you do change it, make sure the EXPLAIN looks something like this:
```
aaxelb: For the old query (with `filter(normalizeddata_isnull=True)`), I get an EXPLAIN pretty close to what you put here:
```
Limit  (cost=0.88..13803.64 rows=363 width=1611)
  ->  Nested Loop  (cost=0.88..13803.64 rows=363 width=1611)
        ->  Nested Loop  (cost=0.74..13740.73 rows=363 width=1195)
              ->  Merge Left Join  (cost=0.44..13609.28 rows=363 width=1163)
                    Merge Cond: (share_rawdatum.id = share_normalizeddata.raw_id)
                    Filter: (share_normalizeddata.id IS NULL)
                    ->  Index Scan using share_rawdata_pkey on share_rawdatum  (cost=0.29..13303.84 rows=72640 width=1163)
                    ->  Index Scan using share_normalizeddata_raw_id_b0718619 on share_normalizeddata  (cost=0.15..102.16 rows=1734 width=8)
              ->  Index Scan using share_sourceuniqueidentifier_pkey on share_sourceuniqueidentifier  (cost=0.29..0.35 rows=1 width=32)
                    Index Cond: (id = share_rawdatum.suid_id)
        ->  Index Scan using share_sourceconfig_pkey on share_sourceconfig  (cost=0.14..0.16 rows=1 width=416)
              Index Cond: (id = share_sourceuniqueidentifier.source_config_id)
```
but for the new query (with `annotate`), I get something different:
```
Limit  (cost=18.35..6510.74 rows=500 width=1611)
  ->  Hash Join  (cost=18.35..471625.27 rows=36320 width=1611)
        Hash Cond: (share_sourceuniqueidentifier.source_config_id = share_sourceconfig.id)
        ->  Merge Join  (cost=0.69..319653.81 rows=36320 width=1195)
              Merge Cond: (share_rawdatum.suid_id = share_sourceuniqueidentifier.id)
              ->  Index Scan using share_rawdatum_suid_id_d6dd4177 on share_rawdatum  (cost=0.29..316583.48 rows=36320 width=1163)
                    Filter: (((NOT no_change) OR (no_change IS NULL)) AND (NOT (alternatives: SubPlan 3 or hashed SubPlan 4)))
                    SubPlan 3
                      ->  Index Only Scan using share_normalizeddata_raw_id_b0718619 on share_normalizeddata u0_2  (cost=0.15..36.31 rows=9 width=0)
                            Index Cond: (raw_id = share_rawdatum.id)
                    SubPlan 4
                      ->  Seq Scan on share_normalizeddata u0_3  (cost=0.00..34.34 rows=1734 width=4)
              ->  Index Scan using share_sourceuniqueidentifier_pkey on share_sourceuniqueidentifier  (cost=0.29..2460.89 rows=72640 width=32)
        ->  Hash  (cost=15.07..15.07 rows=207 width=416)
              ->  Seq Scan on share_sourceconfig  (cost=0.00..15.07 rows=207 width=416)
        SubPlan 1
          ->  Index Only Scan using share_normalizeddata_raw_id_b0718619 on share_normalizeddata u0  (cost=0.15..36.31 rows=9 width=0)
                Index Cond: (raw_id = share_rawdatum.id)
        SubPlan 2
          ->  Seq Scan on share_normalizeddata u0_1  (cost=0.00..34.34 rows=1734 width=4)
```
Is there something odd about my data, or does the EXPLAIN in this comment not line up with the query?

share/tasks.py
```diff
@@ -44,7 +44,12 @@ def transform(self, raw_id):
         graph = transformer.transform(raw)
 
         if not graph or not graph['@graph']:
-            logger.warning('Graph was empty for %s, skipping...', raw)
+            if not raw.normalizeddata_set.exists():
+                logger.warning('Graph was empty for %s, setting no_change to True', raw)
+                RawDatum.objects.filter(id=raw_id).update(no_change=True)
```
aaxelb: Transformers don't compare against the database, so this'll only be hit when the raw datum contains nothing the transformer even recognizes, but not in a way that makes it fail... I think this check belongs where we [save changesets](https://github.com/CenterForOpenScience/SHARE/blob/develop/share/models/change.py#L30)

share/models/ingest.py
```diff
@@ -380,6 +381,13 @@ class RawDatum(models.Model):
     date_modified = models.DateTimeField(auto_now=True, editable=False)
     date_created = models.DateTimeField(auto_now_add=True, editable=False)
 
+    no_change = models.NullBooleanField(null=True, help_text=(
+        'Indicates that this RawDatum does not contain any new information. '
+        'This allows the RawDataJanitor to find records that have not been processed.'
+        'Records that do not contain changes will not have a NormalizedData associated with them, '
```
aaxelb: I think this is wrong... Records that do not contain changes *will* have a NormalizedData, but will not have a ChangeSet.

share/janitor/tasks.py
chrisseto: I trimmed the original explain to remove the parts that aren't super important.
Really the only important part is `Nested Loop Anti Join`.
Postgres' query planner will decide the optimal way to perform a query based on the available hardware, data size, etc, etc. Given that you only have ~500 rows a join is probably faster.

share/tasks.py
chrisseto: This is the only time that rawdata could be processed and not generate a normalized data, the case that this boolean exists to handle.
If there is not a way to detect this and we ever have more dud rawdata than `limit` this task will no longer do anything.

My original explanation is definitely wrong, however, as a normalized data would still be created in that case there is no need for that flag (currently, it would be very helpful).

share/janitor/tasks.py
chrisseto: You're right but I think I would prefer the ability to turn logging on rather than have this be slightly faster. The additional joins don't add all that much overhead, it's the anti-join (and the order by) that brought the query to a crawl.

share/models/ingest.py
chrisseto: Better?
```python
        'Indicates that this RawDatum resulted in an empty graph when transformed. '
        'This allows the RawDataJanitor to find records that have not been processed.'
        'Records that result in an empty graph will not have a NormalizedData associated with them, '
        'which would otherwise look like data that has not yet been processed.'
```

Any ideas for a better name than `no_change`?

share/tasks.py
aaxelb: Oh right, that makes sense. I was hung up on "no change" rather than thinking about the janitor. 

Would it make more sense for the janitor to look for raws without IngestJobs (once those are a thing)?

share/models/ingest.py
aaxelb: Yeah, much better.

`no_output`? `empty_graph`? `full_of_nothing`? `pointless`?

chrisseto: I'm digging `no_output`?

share/harvest/scheduler.py
```diff
@@ -34,12 +34,12 @@ def all(self, cutoff=None, allow_full_harvest=True, **kwargs):
         if hasattr(self.source_config, 'latest'):
             latest_date = self.source_config.latest
         else:
-            latest_date = HarvestLog.objects.filter(source_config=self.source_config).aggregate(models.Max('end_date'))
+            latest_date = HarvestLog.objects.filter(source_config=self.source_config).aggregate(models.Max('end_date'))['end_date__max']
```
aaxelb: Since you did it below, what about `self.source_config.harvest_logs.aggregate...`?

tests/share/test_harvest.py
```diff
@@ -401,3 +401,45 @@ def test_harvest_after(self, monkeypatch, now, end_date, harvest_after, should_r
...
+            end_date=pendulum.parse('2017-01-02').date(),
+        )
+
+        factories.HarvestLogFactory(
+            source_config=source_config,
+            start_date=pendulum.parse('2018-01-01').date(),
+            end_date=pendulum.parse('2018-01-02').date(),
+        )
+
+        assert len(HarvestScheduler(source_config).all(cutoff=pendulum.parse('2018-01-01').date())) == 0
```
aaxelb: I realize this isn't new, but it seems odd (or at least unintuitive) that `all()` doesn't fill in any gaps, it just goes from the most recent HarvestLog until the cutoff. I don't know that I would rename it (because I can't think of a better name), but I think it'd be worth mentioning that in the docstring.

chrisseto: It will fill in gaps, just not if earliest date has a harvestlog. Which is just a really bad way to detect if all the other logs have been made... We should probably just toss a boolean field onto SourceConfig to keep track of that.

chrisseto:
Merged via https://github.com/CenterForOpenScience/SHARE/releases/tag/2.9.7

PR #696, opened 2017-07-05
[Feature] Defrankenize

chrisseto:


share/models/creative.py
```diff
@@ -1,16 +1,23 @@
...
 from share.models.fields import ShareManyToManyField, ShareURLField
 
 from share.util import strip_whitespace, ModelGenerator
 
 
+logger = logging.getLogger(__name__)
+
 # Base Creative Work class
 
+
```
laurenbarker: Probably both of the newlines between the comment and the class should be deleted.

share/models/creative.py
```diff
@@ -43,6 +50,32 @@ class Meta(ShareObject.Meta):
...
+                    continue
+
+                logger.warning('Removing all %s', field.related_name)
+                relation = getattr(self, field.get_accessor_name())
+                num_deleted, stats = Change.objects.filter(id__in=relation.values_list('change_id', flat=True)).delete()
+                logger.warning('Deleted %d changes to remove %s', num_deleted, field.related_name)
+
+                assert num_deleted == stats.pop('share.Change', 0)
+
+                if stats:
```
laurenbarker: Will this ever be falsey if `assert num_deleted == stats.pop('share.Change', 0)` is True?

share/models/creative.py
```diff
@@ -43,6 +50,32 @@ class Meta(ShareObject.Meta):
...
+                    logger.error('Unexpectedly removed other rows, %r', stats)
+                    raise ValueError('Unexpectedly removed other rows, {!r}'.format(stats))
+
+            logger.info('Relations removed')
+            self.administrative_change(is_deleted=True, title='Defrankenized work')
```
laurenbarker: Would it be better to set a flag on the model instead of changing the title?

tests/share/models/factories.py
```diff
@@ -1,117 +1,14 @@
...
-
-
-class AgentWorkRelationFactory(TypedShareObjectFactory):
-    cited_as = factory.Faker('name')
-    agent = factory.SubFactory(AgentFactory)
-    creative_work = factory.SubFactory(AbstractCreativeWorkFactory)
-
-    class Meta:
-        model = models.AgentWorkRelation
+AgentFactory = AbstractAgentFactory
```
laurenbarker: Why was agent the only thing left in here?

tests/share/models/test_shareobject.py
```diff
@@ -169,3 +170,62 @@ def test_related_objects(self):
...
+class TestDefrankenize:
+
+    def test_must_be_really_sure_about_it(self):
+        work = factories.AbstractCreativeWorkFactory()
+
+        with pytest.raises(ValueError) as e:
+            work.defrankenize()
+        assert e.value.args == ('You have to be really sure about this', )
+
+        with pytest.raises(ValueError) as e:
```
laurenbarker: Does this need to happen twice?

tests/share/models/test_shareobject.py
```diff
@@ -169,3 +170,62 @@ def test_related_objects(self):
...
+            work.defrankenize(True)
+        assert e.value.args == ('You have to be really sure about this', )
+
+        work.defrankenize(im_really_sure_about_this=True)
+
+    def test_resets_name(self):
+        work = factories.AbstractCreativeWorkFactory(title='Not a franken', is_deleted=False)
```
laurenbarker: This may be sarcastic but the 'Not a fraken' as a title is a little misleading for this test.

tests/share/models/factories.py
chrisseto: It's the only class that had a non-standard name and would only be used from this module.
Everything else is replaced by the above imports.

tests/share/models/test_shareobject.py
chrisseto: It's two different test cases

laurenbarker: Oh I missed the `True` there 👍

share/models/creative.py
chrisseto: It would be, but I'd rather not deal with the migration and I like to think SSO will be merged soon enough that this won't be a problem.

laurenbarker:
@chrisseto, looks good to me 🥇

PR #699, opened 2017-07-11
Feature/search indexing

chrisseto:


tests/share/search/test_indexing.py
```diff
@@ -0,0 +1,370 @@
...
+            self.message_factory(ids=[3, 4, 5]),
+            self.message_factory(ids=[6, 7, 8, 9]),
+        ]
+
+        flattener = indexing.MessageFlattener(messages)
+        assert flattener.pending == []
+
+        # -1 as a static offset (starts with 2 1s)
+        # +2 to account for the 2 empty messages at the begining of the list
+        lengths = [i - 1 + 2 for i in range(2, 6) for _ in range(i)]
```
aaxelb: I appreciate the logic, but this is way more opaque than just writing out a list of the expected lengths, and would require more thought to update if the test case ever changes. It also makes a longer list than necessary (extra 6s on the end), which could be confusing.
```
lengths = [3, 3, 4, 4, 4, 5, 5, 5, 5, 6]
```

tests/share/search/test_indexing.py
```diff
@@ -0,0 +1,370 @@
...
```
aaxelb: Ditto

tests/share/search/test_indexing.py
```diff
@@ -0,0 +1,370 @@
...
```
aaxelb: Tritto

share/search/indexing.py
```diff
@@ -0,0 +1,298 @@
...
+        self.pending.clear()
+
+    def __len__(self):
+        return sum([len(msg) for msg in self.messages], 0)
+
+    def __iter__(self):
+        return self
+
+    def __next__(self):
+        if self.current is None:
```
aaxelb: I suppose it does make those list comprehensions in the tests more mathematical, but I don't see any other reason for it...

chrisseto: It forces a "look ahead" so that we're always one ahead of the internal buffer. So as soon as the final element of a list is consumed, it will be pushed into pending rather than on the next call. 

I moved the initial call into `__init__` because that make more sense 👍

aaxelb: Ah, makes sense :1st_place_medal:

PR #702, opened 2017-07-13
[Improvement][SHARE-770] Use Django-Include to speed up the OAI-PMH e…

chrisseto:
# Before

screen shot 2017-07-13 at 12 58 43 pm

# After

screen shot 2017-07-13 at 12 57 56 pm

### DIFF
--- a/share/oaipmh/repository.py
+++ b/share/oaipmh/repository.py
@@ -104,7 +104,9 @@ class OAIRepository:
         if not queryset.exists():
             self.errors.append(oai_errors.NoResults())
             return [], None, None
-        # TODO use django-include to prefetch relations/agents/identifiers/etc. in one query
+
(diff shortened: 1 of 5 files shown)


share/oaipmh/repository.py
```diff
@@ -104,7 +104,9 @@ def _load_page(self, kwargs):
         if not queryset.exists():
             self.errors.append(oai_errors.NoResults())
             return [], None, None
-        # TODO use django-include to prefetch relations/agents/identifiers/etc. in one query
+
+        queryset = queryset.include('identifiers', 'subjects', 'sources__source', 'incoming_creative_work_relations', 'agent_relations__agent')
```
aaxelb: Would include understand `related_agents` instead of `agent_relations__agent`?

chrisseto: Yes, but it would not include the relations; only the agents would be present.

PR #704, opened 2017-07-25
[SHARE-832][feature]Add Symbiota harvester (Additional dev.)

sheriefvt:
This is a follow-up work for a previous merged PR #671 
* Fix contact info (name) for most of the cases.
* Add external identifier (URL)
* Update source .yaml


share/transformers/org_swbiodiversity.py
```diff
@@ -90,10 +93,25 @@ def unwrap_data(self, input_data):
...
                     contact_name = contact.split('(', 1)[0].strip()
-                    if ', Curator' in contact_name:
-                        contact_name = contact_name.replace(', Curator', '').strip()
-                    if contact and contact_email and re.match(r"[^@]+@[^@]+\.[^@]+", contact_email):
+                    remove_list = ['Curator', 'Science Division Chair', 'Curator', 'Collections Manager',
+                                   'Collections Manager', 'Administrative Director', 'Director', 'Director and Curator',
+                                   'Mycologist and Director', ',']
+                    for item in remove_list:
+                        if item.lower() in contact_name.lower():
+                            contact_name = contact_name.lower().replace(item.lower(), '').strip()
```
aaxelb: This means `contact_name` will be lower case if anything is replaced. Could you use a case-insensitive regex instead?

share/transformers/org_swbiodiversity.py
```diff
@@ -125,6 +143,15 @@ def unwrap_data(self, input_data):
             collection_statistics = start.find_all_next('li')
             collection_statistics = list(map(self.extract_text, collection_statistics))
             data['collection-statistics'] = self.process_collection_stat(collection_statistics)
+
+            link = record.find_next('div', style="margin:3px;")
```
aaxelb: Instead of all this (and instead of `identifiers = ...` above), you could yield the full URL as the identifier [from the harvester](https://github.com/sheriefvt/SHARE/blob/e7eb3f0b8aea33d88ff06aba71a2c35032be5319/share/harvesters/org_swbiodiversity.py#L64) (`yield collection_page.url, str(record)`). It will automatically be added as an identifier.

share/transformers/org_swbiodiversity.py
```diff
@@ -90,10 +93,25 @@ def unwrap_data(self, input_data):
                     contact_dict = {}
                     contact = entry.replace('Contact:', '').strip()
                     contact_email = contact[contact.find("(") + 1:contact.find(")")]
+                    if re.match(r"[^@]+@[^@]+\.[^@]+", contact_email):
```
aaxelb: Also, single quotes, please

sheriefvt:
@aaxelb, thanks for your feedback. YAML has a feature called 'anchors', which let you easily duplicate content across your yaml document. Both of transformer_kwargs and  harvester_kwargs will have the same url value. I used this to avoid duplication and to avoid updating url in two places if need to be updated. Since we are using the yielded identifier from the harvester, it is not needed now and removed.

aaxelb:
@sheriefvt Oh neat, I didn't know about yaml anchors. That's handy.

PR #705, opened 2017-07-26
[SHARE-828][Feature]Update deprecated regex to support Python 3.6

sheriefvt:
## Ticket
https://openscience.atlassian.net/browse/SHARE-828

## Problem
Some of our regexes use a deprecated syntax that breaks in python 3.6
Deprecations can be found by running python -Wa manage.py
We need to upgrade to python 3.6

## Solution:

- Remove \B from regex for both ISNI and DOI.
- Run **test_isni_link** and **test_doi_link** under tests/share/normalize/test_links and make sure all tests pass.

### DIFF
--- a/share/transform/chain/links.py
+++ b/share/transform/chain/links.py
@@ -672,7 +672,7 @@ class ISNILink(AbstractIRILink):
         (None, 150000007),
         (350000001, None),
     )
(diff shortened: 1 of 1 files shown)


aaxelb:
Looks good to me 🐝

PR #701, opened 2017-07-13
Update Django

sloria:
https://docs.djangoproject.com/en/1.11/releases/1.11.2/
https://docs.djangoproject.com/en/1.11/releases/1.11.3/

PR #709, opened 2017-08-08
Update django 1.11.4

sloria:
Same as #701 , just a little more.

### DIFF
--- a/share/models/base.py
+++ b/share/models/base.py
@@ -58,6 +58,10 @@ class ShareObjectMeta(ModelBase):
 
     def __new__(cls, name, bases, attrs):
         if (models.Model in bases and attrs['Meta'].abstract) or len(bases) > 1:
+            attrs = {
+                **{k: v() for k, v in cls.share_attrs.items()},
+                **attrs
+            }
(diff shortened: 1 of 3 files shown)


share/models/base.py
```diff
@@ -58,6 +58,10 @@ class ShareObjectMeta(ModelBase):
 
     def __new__(cls, name, bases, attrs):
         if (models.Model in bases and attrs['Meta'].abstract) or len(bases) > 1:
+            attrs = {
```
chrisseto: How does this differ from lines, 89, 90, and 101?

aaxelb: This adds the "base class" to abstract models, too. A [bugfix in 1.11.3](https://code.djangoproject.com/ticket/28282) generates names for all indexes whenever a model is defined. `ShareObject`'s [index on `date_created`](https://github.com/CenterForOpenScience/SHARE/blob/9d8d0d2294718a6ff8f5fba5dded421951770926/share/models/base.py#L158) would cause an error because there was no `ShareObject.date_created` field. This seemed like the simplest solution.

PR #711, opened 2017-08-10
[SHARE-961][Improvement] Add 409 on conflict and PUT to source api route

alexschiller:
## Purpose
There is no way to retrieve data back from the API or to change source data.

## Changes
On conflict this returns the serialized data instead of failing. Adds a put route to update icons and homepage data.

## Side Effects
This is bad. Really not good. The way the API is already set up does not take advantage of serializers and is very un-restful. I attempted to make it better, but everything failed and I defaulted back to this bad way of doing things. It does its job but can't say more than that. 

Also, permissions for users may need to allow change to make `PUT` work. 

## Ticket
https://openscience.atlassian.net/browse/SHARE-961

### DIFF
--- a/api/views/workflow.py
+++ b/api/views/workflow.py
@@ -56,6 +56,12 @@ class SourceViewSet(viewsets.ReadOnlyModelViewSet):
     def get_queryset(self):
         return Source.objects.exclude(icon='').exclude(is_deleted=True)
 
+    def get_object(self, id):
+        try:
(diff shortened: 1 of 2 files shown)


chrisseto:
https://github.com/CenterForOpenScience/SHARE/blob/develop/CONTRIBUTING.md#contributing

aaxelb:
Pulled this into #719, thanks @alexschiller!

PR #712, opened 2017-08-10
[SHARE-962][Improvement] Don't index works that have 0 identifiers

chrisseto:


chrisseto:
@aaxelb, excellent question. I forgot to add the script that does that.

share/management/commands/removeanonymousworks.py
```diff
@@ -0,0 +1,26 @@
...
+from project import celery_app
+
+from share.models import WorkIdentifier
+from share.models import AbstractCreativeWork
+
+
+class Command(BaseCommand):
+
+    def handle(self, *args, **options):
+        qs = AbstractCreativeWork.objects.annotate(
```
aaxelb: Is annotate + exclude better than just `.filter(identifiers=None)` or `.filter(identifiers__isnull=True)`?

chrisseto: I believe so, postgres' `Exists` query's are generally faster than the alternative join that django would generate

PR #713, opened 2017-08-14
[SHARE-932][Improvement] API clean up

chrisseto:


aaxelb:
Pulled this into #719, thanks @chrisseto!

PR #715, opened 2017-10-12
[SHARE-978][Feature] Add `blocked_sets` option to OAI/MODS transformers

aaxelb:
Allow adding `blocked_sets` in a source config's `transformer_kwargs` for `oai` and `mods` transformers to set a blacklist of set specs. If an incoming datum has a blacklistec set spec, discard it.

https://openscience.atlassian.net/browse/SHARE-978


tests/share/transformers/test_oai_dc.py
```diff
@@ -0,0 +1,72 @@
...
+      <dc:date>2017-09-11T03:13:21Z</dc:date>
+      <dc:type>article</dc:type>
+      <dc:identifier>http://ro.uow.edu.au/aiimpapers/2062</dc:identifier>
+      <dc:identifier>oai://ro.uow.edu.au/aiimpapers-3064</dc:identifier>
+    </oai_dc:dc>
+  </metadata>
+</record>
+        '''.strip()
+
+    @pytest.mark.parametrize('approved_sets, blocked_sets, allowed', [
```
sheriefvt: what is the difference between allowed and approved

tests/share/transformers/test_oai_dc.py
```diff
@@ -0,0 +1,72 @@
...
+    def datum(self):
+        return '''
+<record xmlns="http://www.openarchives.org/OAI/2.0/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
+  <header>
+    <identifier>oai:share.osf.io:461B6-A5C-956</identifier>
+    <datestamp>2017-09-12T13:35:17Z</datestamp>
+    <setSpec>au.uow</setSpec>
+  </header>
+  <metadata>
+    <oai_dc:dc xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:oai_dc="http://www.openarchives.org/OAI/2.0/oai_dc/" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://www.openarchives.org/OAI/2.0/oai_dc/ http://www.openarchives.org/OAI/2.0/oai_dc.xsd">
```
sheriefvt: is this real names/titles of authors/papers?.. would this test appear in search engines. In this case would it be better to use fake name/title

share/transform/chain/utils.py
```diff
@@ -128,3 +129,18 @@ def get_emails(s):
...
+def oai_allowed_by_sets(data, blocked_sets=None, approved_sets=None):
+    # TODO do this in the Regulator, in a ValidationStep
+    blocked_sets = set(blocked_sets or [])
+    approved_sets = set(approved_sets or [])
+    if blocked_sets or approved_sets:
+        set_specs = set(x.replace('publication:', '') for x in etree.fromstring(data).xpath(
+            'ns0:header/ns0:setSpec/node()',
+            namespaces={'ns0': 'http://www.openarchives.org/OAI/2.0/'}
+        ))
+        if (blocked_sets and (set_specs & blocked_sets)) or (approved_sets and not (set_specs & approved_sets)):
```
sheriefvt: if statement too long, can you write an explanation in a comment

share/transformers/mods.py
```diff
@@ -454,18 +464,10 @@ class RootParser(MODSCreativeWork):
...
-                namespaces={'ns0': 'http://www.openarchives.org/OAI/2.0/'}
-            ))
-            if not (specs & set(approved_sets)):
-                logger.warning('Series %s not found in approved_sets for %s', specs, self.config.label)
-                return (None, None)
-
-        return super().do_transform(data)
+    def do_transform(self, datum):
+        if not oai_allowed_by_sets(datum, self.kwargs.get('blocked_sets'), self.kwargs.get('approved_sets')):
+            return (None, None)
```
sheriefvt: can we add a logger warning/message here

share/transformers/oai.py
```diff
@@ -317,15 +317,7 @@ class RootParser(OAICreativeWork):
...
-                namespaces={'ns0': 'http://www.openarchives.org/OAI/2.0/'}
-            ))
-            if not (specs & set(approved_sets)):
-                logger.warning('Series %s not found in approved_sets for %s', specs, self.config.label)
-                return (None, None)
-
-        return super().do_transform(data)
+    def do_transform(self, datum):
+        if not oai_allowed_by_sets(datum, self.kwargs.get('blocked_sets'), self.kwargs.get('approved_sets')):
+            return (None, None)
```
sheriefvt: same here too

tests/share/transformers/test_oai_dc.py
```diff
@@ -0,0 +1,72 @@
...
+        (None, None, True),
+        ([], [], True),
+        (['au.uow'], None, True),
+        (None, ['au.uow'], False),
+        (['uow'], [], False),
+        ([], ['foo', 'bar'], True),
+        (['au.uow'], ['au.uow'], False),
+        (['one', 'two'], ['three', 'four'], False),
+        (['one', 'two', 'au.uow'], ['three', 'four'], True),
+        (['one', 'two'], ['three', 'four', 'au.uow'], False),
```
sheriefvt: nice test cases

tests/share/transformers/test_oai_dc.py
aaxelb: `approved_sets` is the list of set specs that are explicitly allowed (the whitelist). `allowed` is a boolean, whether the work should be let into SHARE in this test. It could be named more clearly...

tests/share/transformers/test_oai_dc.py
aaxelb: Yeah, I just grabbed actual data from staging. You're right, it would be better to use fake data, I'll change it.

share/transformers/mods.py
aaxelb: `oai_allowed_by_sets` logs a warning when it returns `False`, I don't think it'd be different in each transformer.

PR #716, opened 2017-10-13
[SHARE-989][Feature] Add command to enforce set white/blacklists

aaxelb:
Usage:
`python manage.py enforce_set_lists <source_config> [--dry] [--commit] [--whitelist]`

Will delete works that should not have been let in according to a source config's `approved_sets` and `blocked_sets`. If a work also has other sources, just remove the source instead of deleting the work.


share/management/commands/enforce_set_lists.py
```diff
@@ -0,0 +1,95 @@
...
+    source from it instead of deleting.
+
+    By default enforce blacklists only, because enforcing whitelists is slooow.
+    Pass --whitelist to enforce whitelists too.
+    """
+
+    def add_arguments(self, parser):
+        parser.add_argument('source_configs', nargs='+', type=str, help='Labels of the source configs to look at')
+        parser.add_argument('--dry', action='store_true', help='Print changes that would be made, but make no changes')
+        parser.add_argument('--commit', action='store_true', help='If omitted, abort the transaction after making changes')
```
laurenbarker: What does aborting the transaction do? Not actually make the changes?

aaxelb: Yeah, with neither `--dry` nor `--commit`, it'll actually call `administrative_change` and do stuff in the database, then roll it all back. It's useful to check that the changes won't give a database error without actually saving those changes.

laurenbarker:
Once there are tests, this look good to merge ⚡️

PR #719, opened 2017-10-26
[SHARE-932][SHARE-961][Improvement] API refactor, allow updating Sources

aaxelb:
Combines #711 and #713, with additional cleanup.

* Reorganize API code to be more maintainable and consistent
  * Clump view/serializer/urls together for each type (e.g. `api/sources/serializers.py`, `api/sources/views.py`, etc.)
* Fork django-rest-framework-json-api, add some optional configuration to support our wants/needs
  * Allow using a custom ID in the API, rather than hardcoding to `pk`
  * Allow omitting keys with null values, configured per model
  * Allow returning data with error responses
* When trying to create a source with a name/title that already exists, return 409 and include the existing source in the response
* Allow updating sources with PUT/PATCH
  * For now, no object-level permissions, just global "can change source" permission
  * NOTE: Updating a source's `long_title` will NOT update the elastic index automatically


api/shareobjects/serializers.py
```diff
@@ -0,0 +1,13 @@
+from api.base.serializers import ShareSerializer
+
+
+class ShareObjectSerializer(ShareSerializer):
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        # remove hidden fields
```
laurenbarker: Unclear comment

api/sourceregistrations/views.py
```diff
@@ -42,12 +43,3 @@ class ProviderRegistrationViewSet(viewsets.ModelViewSet):
 
     def get_queryset(self):
         return ProviderRegistration.objects.filter(submitted_by_id=self.request.user.pk)
-
-    def create(self, request, *args, **kwargs):
```
laurenbarker: Should this be deleted?

api/urls.py
```diff
@@ -1,137 +1,30 @@
...
-
-model_schema_patterns = [
-    url(r'schema/{}/?'.format(v.MODEL.__name__), v.as_view())
-    for v in views.ModelSchemaView.model_views
-]
+app_name = 'api'
 
 urlpatterns = [
+    url('^$', RootView.as_view()),
+    url('^', include('api.banners.urls')),
```
laurenbarker: Do you want to change these?

api/users/serializers.py
```diff
@@ -0,0 +1,44 @@
...
+
+from rest_framework_json_api import serializers
+
+from share import models
+
+from api.base import ShareSerializer
+
+
+class ShareUserSerializer(ShareSerializer):
+    def __init__(self, *args, token=None, **kwargs):
```
laurenbarker: Remove `token` -Abram

api/users/views.py
```diff
@@ -0,0 +1,25 @@
...
+    """
+    Returns details about the currently logged in user
+    """
+    serializer_class = ShareUserSerializer
+
+    def get_queryset(self):
+        return models.ShareUser.objects.filter(pk=self.request.user.pk)
+
+
+# TODO Remove me
```
laurenbarker: This is probably related to moving things to `users/me`. We should update this comment and reference https://openscience.atlassian.net/browse/SHARE-586.

api/util.py
```diff
@@ -0,0 +1,22 @@
...
+
+class HttpSmartResponsePermanentRedirect(http.HttpResponsePermanentRedirect):
+    status_code = 308
+
+
+def absolute_reverse(view_name, *args, **kwargs):
+    return '{}{}'.format(settings.SHARE_API_URL.rstrip('/'), reverse(view_name, *args, **kwargs))
+
+
+def validation_error_has_code(exc, code):
```
laurenbarker: Do we still need this?

share/models/ingest.py
```diff
@@ -82,10 +82,13 @@ class Source(models.Model):
     canonical = models.BooleanField(default=False, db_index=True)
 
     # TODO replace with Django permissions something something, allow multiple sources per user
```
laurenbarker: It would be great to link a ticket here.

api/sourceregistrations/views.py
aaxelb: Yeah, it doesn't do anything that the default `create` doesn't do

api/urls.py
aaxelb: I'm fine with them for now, I think.

PR #724, opened 2017-11-10
[SHARE-994][Fix] Don't triple-escape atom feed content

aaxelb:
Escaping once is enough. We don't need to do it three times.

https://openscience.atlassian.net/browse/SHARE-994

### DIFF
--- a/api/views/feeds.py
+++ b/api/views/feeds.py
@@ -1,5 +1,5 @@
 
-from xml.sax.saxutils import escape
+from xml.sax.saxutils import unescape
 import datetime
 import json
 import logging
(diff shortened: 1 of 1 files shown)


api/views/feeds.py
```diff
@@ -27,10 +27,12 @@
 RE_XML_ILLEGAL_COMPILED = re.compile(RE_XML_ILLEGAL)
 
 
-def sanitize_for_xml(s):
+def prepare_string(s):
     if s:
         s = RE_XML_ILLEGAL_COMPILED.sub('', s)
-        return escape(s)
+        # Strings will be autoescaped during XML generation, and data from elasticsearch
+        # is already bleached. Unescape to escape extra escapes.
```
sheriefvt: 👀

PR #731, opened 2017-12-14
[SHARE-1002][Improvement] Get all fields from NSFAwards

laurenbarker:
## Purpose
Get all fields from NSFAwards.

## Changes
* Update harvester
* Update transformer

## QA Notes
Quick spot check to make sure things look right with newly harvested NSFAwards works.

Should note that PIs don't show up in the contributor section because the detail page only show people with a "Contributor" or "Creator" relationship. They are returned by the graph endpoint though so could check the network tab in the console.

### DIFF
--- a/share/transformers/gov_nsfawards.py
+++ b/share/transformers/gov_nsfawards.py
@@ -1,10 +1,30 @@
 from share.transform.chain import ctx, ChainTransformer
 from share.transform.chain import links as tools
 from share.transform.chain.parsers import Parser
+from share.transform.chain.utils import format_address
 
 
 def format_url(award_id):
     return 'https://www.nsf.gov/awardsearch/showAward?AWD_ID={}'.format(award_id)
+
+
+def format_org_address(ctx):
+    awardee_address = ctx.get('awardeeAddress', '')
+    awardee_city = ctx.get('awardeeCity', '')
+    awardee_state_code = ctx.get('awardeeStateCode', '')
+    awardee_country_code = ctx.get('awardeeCountryCode', '')
+    awardee_zip_code = ctx.get('awardeeZipCode', '')
+
+    if not any((awardee_address, awardee_city, awardee_state_code, awardee_zip_code, awardee_country_code)):
+        return None
+
+    return format_address(
+        address1=awardee_address,
+        city=awardee_city,
+        state_or_province=awardee_state_code,
+        postal_code=awardee_zip_code,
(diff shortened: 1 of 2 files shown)


share/transformers/gov_nsfawards.py
```diff
@@ -54,54 +67,178 @@ class AffiliatedAgent(Parser):
     )
 
     name = ctx.awardeeName
-    location = tools.Join(tools.Concat(ctx.awardeeCity, tools.Try(ctx.awardeeStateCode)), joiner=', ')
+    location = tools.Join(
+        tools.Concat(ctx.awardeeCity, tools.Try(ctx.awardeeStateCode)),
```
aaxelb: Maybe `awardeeCountryCode`, too?

share/transformers/gov_nsfawards.py
```diff
@@ -54,54 +67,178 @@ class AffiliatedAgent(Parser):
...
     identifiers = tools.Map(tools.Delegate(WorkIdentifier), ctx)
 
     related_agents = tools.Concat(
         tools.Map(tools.Delegate(FunderRelation), ctx),
-        tools.Map(tools.Delegate(ContributorRelation), ctx),
-        tools.Map(tools.Delegate(AgentWorkRelation), tools.Filter(lambda x: 'awardeeName' in x, ctx))
+        tools.Map(tools.Delegate(PIContributorRelation), ctx),
+        tools.Map(
+            tools.Delegate(POContributorRelation),
+            tools.Filter(lambda x: 'poName' in x, ctx)
```
aaxelb: This seems to imply that some records have no program officer, in which case `poName` is omitted instead of set `null`. Is that right? (ditto awardeeName)

laurenbarker: It doesn't create a relation if `poName` isn't there. Is that the question?

aaxelb: I think it will (try to?) create a relation if `poName` is null, though. If a field is irrelevant, do they omit the key or use null?

laurenbarker: They omit keys that aren't there. None of this is documented but I haven't seen any null values for any fields.

aaxelb: I'd be more comfortable with something like `lambda x: x.get('poName')`, just in case that undocumented behavior changes.

PR #734, opened 2017-12-20
[SHARE-1007][Fix] Add x-throttle-token header to OSF harvester

laurenbarker:
*Requires that devops adds the `BYPASS_THROTTLE_TOKEN` as an environment variable on prod. Michael has been notified.* ✅  this has been done

## Purpose
OSF harvest tasks fail, every day. We are getting 429s from the OSF API.

## Changes
Add no throttle header 

## QA notes
Doesn't need QA.

### DIFF
--- a/share/harvesters/io_osf.py
+++ b/share/harvesters/io_osf.py
@@ -18,7 +18,7 @@ class OSFHarvester(BaseHarvester):
     def build_url(self, start_date, end_date, path, query_params):
         # so prod SHARE doesn't get throttled
         if settings.OSF_BYPASS_THROTTLE_TOKEN:
-            self.requests.headers.update({'X-THROTTLE-TOKEN': settings.OSF_BYPASS_THROTTLE_TOKEN})
+            self.session.headers.update({'X-THROTTLE-TOKEN': settings.OSF_BYPASS_THROTTLE_TOKEN})
 
         url = furl(settings.OSF_API_URL + path)
         url.args['page[size]'] = 100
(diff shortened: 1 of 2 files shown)


laurenbarker:
@icereval @binoculars this is the throttle token PR scheduled for v2.14.0 (does not currently have a release date). We will need the throttle token added to the production environment. We don't get throttled on staging so no need to add it there. 

Anything we should be worried about? We have the rate limit for OSF set to 1 request every 3 seconds.

share/harvesters/io_osf.py
```diff
@@ -16,6 +16,10 @@ class OSFHarvester(BaseHarvester):
     VERSION = 1
 
     def build_url(self, start_date, end_date, path, query_params):
+        # so prod SHARE doesn't get throttled
+        if settings.OSF_BYPASS_THROTTLE_TOKEN:
+            self.requests.headers.update({'X-THROTTLE-TOKEN': settings.OSF_BYPASS_THROTTLE_TOKEN})
```
aaxelb: Should probably use `self.session` -- `self.requests` is a rate-limited wrapper around `self.session`, so the `update()` will sort of have a throttle. Probably has no practical effect, but it will create a bunch of unnecessary wrappers as it is.

share/harvesters/io_osf.py
```diff
@@ -16,6 +16,10 @@ class OSFHarvester(BaseHarvester):
     VERSION = 1
 
     def build_url(self, start_date, end_date, path, query_params):
```
aaxelb: Looks like this is based on an old develop, before some things were moved to `future-release/2.15.0`. Could you resolve conflicts?

PR #736, opened 2018-01-05
[SHARE-1027][Feature] Expose Source.canonical field in API

aaxelb:
Allow reading/writing `Source.canonical` at `/api/v2/sources/`

https://openscience.atlassian.net/browse/SHARE-1027


tests/api/test_sources_endpoint.py
```diff
@@ -337,3 +338,30 @@ def test_timeout_url(self, client, source_add_user, mock_icon_urls):
...
+            content_type='application/vnd.api+json',
+            HTTP_AUTHORIZATION=source_add_change_user.authorization(),
+        )
+        assert resp.status_code == 201
+        data_one = flatten_write_response(resp)
+
+        assert data_one['source']['canonical']
+
+        source_url = '{}{}/'.format(self.endpoint, data_one['source']['id'])
+        test_two_data = get_post_body(id=data_one['source']['id'], canonical=False)
```
laurenbarker: So here you're just updating the same source to be not canonical? Could you maybe add a comment or something to make this more clear?

api/sources/serializers.py
```diff
@@ -87,19 +87,19 @@ class Meta(UpdateSourceSerializer.Meta):
...
         label = long_title.replace(' ', '_').lower()
-
-        name = validated_data.get('name', label)
+        name = validated_data.pop('name', label)
 
         with transaction.atomic():
             source, created = models.Source.objects.get_or_create(
                 long_title=long_title,
                 defaults={
-                    'home_page': validated_data.get('home_page', None),
```
laurenbarker: Home page is a required field in the admin app. Would you mind making it not required so we don't have to add a random page if we edit the source? It would be nice if creating sources through the admin functioned the same way as through the API (i.e. not using home page).

Pretty sure it's this form https://github.com/CenterForOpenScience/SHARE/blob/72044b093af7c79f8844aa270395ffbac4557897/share/admin.py#L356

aaxelb: Ohh, I did not know about that form. I'll update that to match the API's current behavior.

PR #768, opened 2018-07-26
[SHARE-1054] Make disambiguation/regulation more extensible

aaxelb:


share/disambiguation.py
```diff
@@ -142,187 +110,98 @@ def _instance_for_node(self, node):
...
-        def __init__(self):
-            self._index = {}
-            self._info_cache = {}
+    def _all(self):
+        try:
+            all = self._node.model.Disambiguation.all
+        except AttributeError:
+            return ()
+        values = tuple((f, v) for f in all for v in self._field_values(f))
+        assert len(values) == len(all), 'Wrong number of values: {} for "all": {}'.format(values, all)
```
chrisseto: `ValueError` would make a bit more sense here, and not mess with pytest.

share/ingest/change_builder.py
```diff
@@ -0,0 +1,255 @@
...
+
+        user = normalized_datum.source  # "source" here is a ShareUser
+        source = user.source
+
+        if disambiguate:
+            instance_map = GraphDisambiguator(source).find_instances(graph)
+        elif instance_map is None:
+            instance_map = {}
+
+        differs = [cls(n, source, instance_map) for n in graph.topologically_sorted()]
```
chrisseto: Maybe `differences`?

share/regulate/regulator.py
```diff
@@ -0,0 +1,119 @@
...
+    VERSION = 1
+
+    def __init__(
+            self,
+            ingest_job=None,
+            source_config=None,
+            regulator_config=None,
+            validate=True,
+    ):
+        assert not ingest_job or not source_config, 'Provider ingest_job or source_config, not both'
```
chrisseto: `ValueError` :+1:

share/regulate/regulator.py
```diff
@@ -0,0 +1,119 @@
...
+        )
+        self._default_steps = Steps(
+            self,
+            regulator_config or settings.SHARE_REGULATOR_CONFIG,
+            validate=validate
+        )
+
+    def regulate(self, graph):
+        try:
+            self._custom_steps.run(graph)
```
chrisseto: Are there any situations where you'd need to run steps after the default steps?

share/regulate/regulator.py
```diff
@@ -0,0 +1,119 @@
...
+            self._run_steps(graph, self.node_steps)
+
+            graph.changed = False
+            self._run_steps(graph, self.graph_steps)
+            if not graph.changed:
+                break
+
+            runs += 1
+            if runs >= self.MAX_RUNS:
+                raise InfiniteRegulationError
```
chrisseto: Might be helpful to include the config that's looping?

share/regulate/steps/__init__.py
```diff
@@ -21,8 +23,14 @@ def error(self, description, node_id=None, exception=None):
         self.logs.append(log)
         raise RegulateError('Regulation failed: {}'.format(description)) from exception
 
+    @abc.abstractmethod
+    def run(self, graph):
+        pass
```
chrisseto: I have very little faith in `abc`, so adding `raise NotImplementedError` makes me feel a little better.

share/regulate/steps/normalize_agent_names.py
```diff
@@ -0,0 +1,97 @@
...
+            part = human[part_name]
+            if part:
+                node[field_name] = part.title()
+
+        node['name'] = ' '.join(filter(None, (
+            node[k] for k in self.NAME_PARTS.values()
+        )))
+
+    def _normalize_non_person(self, node):
+        # TODO reevaluate everything in this method
```
chrisseto: 😠

share/util/__init__.py
```diff
@@ -102,8 +102,8 @@ class CyclicalDependency(Exception):
     pass
 
 
-class TopographicalSorter:
-    """Sort a list of nodes topographically, so a node is always preceded by its dependencies"""
+class TopologicalSorter:
```
chrisseto: 🎉

share/util/graph.py
```diff
@@ -54,7 +77,6 @@ def from_jsonld(cls, nodes):
                     id = v
                 elif k == '@type':
                     type = v
-                    type = v
```
chrisseto: 🤔

share/util/graph.py
```diff
@@ -212,11 +251,41 @@ def named_in_edges(self, to_id):
...
+        """Merge a nodes attrs and edges into another node.
+
+        `from_node` will be deleted.
+        """
+        self.changed = True
+
+        from_model = from_node.model
+        into_model = into_node.model
+
+        assert from_model._meta.concrete_model is into_model._meta.concrete_model, 'Cannot merge nodes of different types'
```
chrisseto: `ValueError`?

share/util/graph.py
```diff
@@ -212,11 +251,41 @@ def named_in_edges(self, to_id):
...
+
+        if choose_winner and len(from_model.__mro__) >= len(into_model.__mro__):
```
chrisseto: Maybe a better name than `choose_winner`?

share/disambiguation.py
aaxelb: Yeah that makes sense. Kinda knew all those assertions were bad but did them anyway...

share/ingest/change_builder.py
aaxelb: Oh whoops, ChangeBuilder used to be called NodeDiffer, because that felt like a good idea at the time. Should probably name this `builders` or something like that.

share/regulate/regulator.py
aaxelb: Not that I've run into.

I think of custom steps as a place for source-specific logic that makes more sense expressed imperatively. Right now that's what the `RunPython` chain links are for, but moving it here lets the transformers can remain declarative. The transformers just allocate information into buckets, then source-specific regulator steps can do additional processing to get it ready for actual regulation. I think it helps to guarantee the transformer's output goes directly into custom regulation, without anything in between.

PR #770, opened 2018-12-13
Perf improvements

aaxelb:
Investigating some memory/performance issues locally yielded a few improvements that might help:
- Faster deduping algorithm for large input graphs
- Skip parsing/deduping unrealistically long names (>200 characters), seems that `nameparser` sometimes chokes

Also allow the `forceingest` script more retries before giving up.

### DIFF
--- a/share/management/commands/forceingest.py
+++ b/share/management/commands/forceingest.py
@@ -12,7 +12,7 @@ from share.tasks.jobs import IngestJobConsumer
 from share.util.osf import osf_sources
 
 
-MAX_RETRIES = 10
+MAX_RETRIES = 128
 GRAVEYARD_IDENTIFIERS = [
     'http://osf.io/8bg7d/',  # prod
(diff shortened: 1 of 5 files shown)


share/management/commands/forceingest.py
```diff
@@ -12,7 +12,7 @@
 from share.util.osf import osf_sources
 
 
-MAX_RETRIES = 10
+MAX_RETRIES = 128
```
chrisseto: 🤔

aaxelb: More accurate would be `MAX_MERGES` -- it'll merge one thing, then retry, then merge one more thing, then retry... if there are 30 contributors that need merging, want to handle it.

Also this is now an interactive script that I babysit, so easy to skip if it looks wrong.

chrisseto: That makes more sense 👍

PR #787, opened 2021-07-07
[ENG-2100] docs: delete old, add new

aaxelb:


ARCHITECTURE.md
```diff
@@ -0,0 +1,76 @@
...
+`share/oaipmh/` describes the [OAI-PMH](https://www.openarchives.org/OAI/openarchivesprotocol.html)
+view for harvesting metadata from SHARE/Trove in bulk.
+
+`api/` describes a mostly REST-ful API that's useful for inspecting records for
+a specific item of interest.
+
+### Testing
+
+`tests/` are tests.
+
```
felliott, line 59: Things that are undocumented but maybe should be:

`share/admin/` => "Django-app for administrative access to SHARE pipeline / db"

`share/ingest/` => ???

`share/janitor/` => ???

`osf_oauth2_adapter/` =>  Django app to support logging in to SHARE via OSF

`bots/elasticsearch/` => ???

`db/` => default Pg database configurations (I think?)

`project/` => Django app that dispatches to other django apps in the repo (I think?)

`Vagrantfile` => Is this deletable?

`bin/share`, `bootstrap.sh`, `up.sh`, `share/bin/*` => files behind sharectl and initial setup.  Some (bootstrap.sh) may be deletable?  Maybe this can be put off until share cmd consolidation.

felliott (changes requested):
Hey @aaxelb!

This is looking good so far!  My comments come in four categories:

1. I'd like to see some diagrams (ascii or bitmap) that show the various processes running within share and how they interact.  There's a lot going on here and a visual map would be helpful.
2. I've pointed out some bits of the old deleted docs that might be worth saving.  I'm not wedded to any of those.  They just jumped out at me while I was skimming the code.  If they should be saved, I don't think they need to be formally integrated right now.  They could just be stuck in a notes header of the existing docs or something like that.
3. Minor updates to the new docs for things to be expanded on.
4. Things I think may have been overlooked in the code architecture doc.  Some of them may be minor enough to be ignored, but I feel like others (e.g. distinct Django apps) should be called out.

Cheers,
@felliott

aaxelb:
thanks @felliott! addressed many of your comments and discussed out-of-band -- going to merge this while acknowledging there are still improvements could be made and gaps could be filled

PR #805, opened 2023-07-14
[wip] iri-focused ingest and search

aaxelb:
todo: describe

(replaces/combines/extends #798 and #804 )


aaxelb:
replaced by #806

PR #818, opened 2024-05-20
simplest simple json

aaxelb:
- add `simple_json` renderer for search api responses
  - includes the osfmap metadata record for each search result, total result count, and pagination links
  - omits everything else
- use `simple_json` renderer for requests with query param `acceptMediatype=application/json` or header `Accept: application/json`


brianjgeiger (approved):
Looks good, near as I can tell. 🐧

PR #821, opened 2024-07-29
[ENG-6052] | Fixes to share for M chips discovered during local setup

sh-andriy:


sh-andriy:
Were encountering errors when running this command `docker-compose up -d elastic8`, such as network problems (share_network helped), still had other issues and after specifying platform it removed incompatible platforms warnings (arm with amd) and set elastic8 up without troubles.

When was testing the `share` project in `osf`  (here http://localhost:5000/search) could not see any of my projects and encountered errors this fix helped me to solve them:
in **docker-compose.ym**l:

in **elastic8** service under environment category removed this line:
`      - cluster.initial_master_nodes=singlenode
`add this line instead:
`      - discovery.type=single-node
`

sh-andriy:
Were encountering errors when running this command `docker-compose up -d elastic8`, such as network problems (share_network helped), still had other issues and after specifying platform it removed incompatible platforms warnings (arm with amd) and set elastic8 up without troubles.

When was testing the `share` project in `osf`  (here http://localhost:5000/search) could not see any of my projects and encountered errors this fix helped me to solve them:
in **docker-compose.ym**l:

in **elastic8** service under environment category removed this line:
`      - cluster.initial_master_nodes=singlenode
`
add this line instead:
`      - discovery.type=single-node
`

mfraezz (approved):
Networking changes should be fine, and the change from `cluster.initial_master_node` to `discovery.type` makes sense -- if we're only running a one-node cluster, it should elect itself whenever it spins up.

I personally ran into an issue spinning up `rabbitmq:management` when it attempted to enable the `stream_sac_coordinator_unblock_group` feature flag, a recent addition in version 3.13. I found that specifying the `3.12-management-alpine` tag instead allowed it to run. This may be a hardware issue, as I found reports of other M-series apple chips encountering the same or similar issues.

Overall, LGTM, but I'll leave it for @aaxelb to confirm and merge. :octocat:

aaxelb (approved):
looks good, thanks! (and test failure unrelated)

PR #825, opened 2024-09-23
[ENG-6265] supplementary metadata

aaxelb:
allow "supplementary" metadata records

- declared by adding a `is_supplementary` query param (regardless of value) to a POST to `/trove/ingest`
- added to all pre-existing index-cards from the same user with the same `focus_iri` (if any)
- will not get an index-card of its own
- will not affect the last-updated timestamp (e.g. in OAI-PMH) of the index-cards it supplements (<== this is the main point -- allows indexing monthly-computed metrics and system-specific administrative metadata without triggering full re-harvests every month)


brianjgeiger:
I'm not seeing a CR on this. Do we need to accelerate getting you help on SHARE? Getting folks doing CR might be a good way to start.

aaxelb:
@brianjgeiger it was reviewed by @felliott in a call we had together, going thru each change and discussing what/why -- agree doing review together is a good way to share share knowledge, but will be sure to leave a comment (or harangue for actual github approval) when doing that in the future

brianjgeiger:
@aaxelb Awesome, thank you. And thank you @felliott for the assistance.

PR #826, opened 2024-10-15
[ENG-6313] allow raw data to expire

aaxelb:
- add `RawDatum.expiration_date` model field (with index)
- add "expiration_date" query param to `/trove/ingest`
- update digestive tract to appropriately handle expired data
- add periodic task to expel expired data daily
  - if a supplementary record's latest RawDatum expires, delete the `SupplementaryIndexcardRdf` and send affected indexcards back thru `digestive_tract.derive`
  - if an indexcard's latest non-supplementary RawDatum expires, mark that indexcard deleted (using existing mechanisms)


mfraezz (approved):
LGTM ⭐.
Legible, well-tested, no obvious bugs.

PR #824, opened 2024-08-26
[wip/analysis][ENG-6075] trove-search performance analysis

aaxelb:
ponder ways to improve performance for `/trove/index-card-search` and `/trove/index-value-search` requests -- see `_TODO.txt` for notes (all code changes are exploratory)

PR #828, opened 2024-10-25
[ENG-6189] trovesearch denormalized

aaxelb:
- add `trovesearch_denorm` index strategy to speed up `/trove/index-card-search` and `/trove/index-value-search` requests
  - used by default when the `TROVESEARCH_DENORMILY` feature flag is up
  - stop using `nested` fields (what a wild choice that was)
    - to support indexing values at arbitrary paths, use dynamic templates
    - to support value-search with `cardSearchFilter`/`cardSearchText` params, index additional docs for each iri value in an indexcard (each with a full copy of the card's filterable fields -- this is the denormalization part)
  - include integer values (for sorting only, so far)
- allow sorting by property-path (not only by a single property)

and assorted improvements from meanderings in #824
- move the definition of available index-strategies from settings to static code (to fix having code that depends on settings being just so)
- replace static methods on `IndexStrategy` with functions in `share.search.index_strategy`
- add admin view to view mappings for an index
- update abstract `Elastic8IndexStrategy` to support indexing multiple docs per index-card (required for denormalization above)
- make tests reusable for index-strategies supporting trove-search
- consolidate some logic shared across trove-y index strategies to reusable utils in `share.search.index_strategy._trovesearch_util`
- more accurate type annotations


share/search/index_strategy/trovesearch_denorm.py
```diff
@@ -0,0 +1,985 @@
...
+        # delete all per-value docs (to account for missing values)
+        self.es8_client.delete_by_query(
+            index=list(indexnames),
+            query={'bool': {'must': [
+                {'terms': {'card.card_pk': messages_chunk.target_ids_chunk}},
+                {'exists': {'field': 'iri_value.value_iri'}},
+            ]}},
+        )
+        # (possible optimization: instead, hold onto doc_ids and (in `after_chunk`?)
+        #                         delete_by_query excluding those)
```
mfraezz, line 159: For my own understanding, can you explain what/why this is doing? When messages come in chunked (e.g. for an index backfill), this will delete existing PKs from that incoming chunk if there's a `value_iri`?

share/search/index_strategy/trovesearch_denorm.py
```diff
@@ -0,0 +1,985 @@
...
+            )
+            if settings.DEBUG:
+                logger.info(json.dumps(_query, indent=2))
+            try:
+                _es8_response = self.index_strategy.es8_client.search(
+                    **_query,
+                    index=self.indexname,
+                )
+            except elasticsearch8.TransportError as error:
+                raise exceptions.IndexStrategyError() from error  # TODO: error messaging
```
mfraezz, line 234: Minor: how annoying would it be to add error messaging instead of the `TODOs` here and above?

share/search/index_strategy/trovesearch_denorm.py
```diff
@@ -0,0 +1,985 @@
...
+        self,
+        valuesearch_params: ValuesearchParams,
+        es8_response: dict,
+        cursor: _SimpleCursor,
+    ) -> ValuesearchResponse:
+        _iri_aggs = es8_response['aggregations'].get('agg_valuesearch_iris')
+        if _iri_aggs:
+            _buckets = _iri_aggs['buckets']
+            _bucket_count = len(_buckets)
+            # WARNING: terribly inefficient pagination (part two)
```
mfraezz, line 398: Any guesses as to how terribly-inefficient -- worth benchmarking? It looks fairly similar to the `FlatsIndexStrat`'s pagination

share/search/index_strategy/trovesearch_denorm.py
```diff
@@ -0,0 +1,985 @@
...
+        (_query,) = queries
+        return _query
+    return {'bool': {'should': list(queries), 'minimum_should_match': 1}}
+
+
+###
+# cursor implementations
+
+@dataclasses.dataclass
+class _SimpleCursor:
```
mfraezz, line 892: Some of the other `dataclasses` in here are very-similar-but-somewhat-different-from some of the `dataclasses` found in `indexcard_flats`, but it looks like the only change in this one as compared to `index_card_flats._SimpleCursor` is the `assert` in `from_page_param`. 

Should this (and/or others, potentially) be DRY'd up a bit?

mfraezz (commented):
Generally looks pretty good. A couple questions and some potential tidying, but no obvious bugs. Good test coverage.

Pass complete :octocat:

project/settings.py
```diff
@@ -314,52 +314,15 @@ def split(string, delim):
...
-    ELASTICSEARCH8_SECRET = os.environ.get('ELASTICSEARCH8_SECRET')
-    ELASTICSEARCH8_CLUSTER_SETTINGS = {
-        'URL': ELASTICSEARCH8_URL,
-        'AUTH': (
-            (ELASTICSEARCH8_USERNAME, ELASTICSEARCH8_SECRET)
-            if ELASTICSEARCH8_SECRET is not None
-            else None
-        ),
-        'CERT_PATH': ELASTICSEARCH8_CERT_PATH,
-    }
```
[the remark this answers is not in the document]
aaxelb: not anymore -- one of the cleanup changes i kept from #824 was removing the index-strategy config from django settings and instead making (for example) the base elastic8 strategy use `settings.ELASTICSEARCH8_URL` directly

will be worth a fresh rethink if index strategies are ever abstracted into a reusable form, but as it is seemed worse the more strategies i added

tests/share/search/index_strategy/test_strategy_selection.py
```diff
@@ -34,36 +42,29 @@ def test_all_strategies(self, mock_elastic_clients, expected_strategy_classes):
...
                 assert specific_index.index_strategy is index_strategy
-            # bad calls:
-            with pytest.raises(IndexStrategyError):
-                IndexStrategy.get_for_sharev2_search('bad-request')
-            with pytest.raises(IndexStrategyError):
-                IndexStrategy.get_for_sharev2_search()
-            with pytest.raises(IndexStrategyError):
-                IndexStrategy.get_for_sharev2_search(requested_name=None)
+        with pytest.raises(IndexStrategyError):
+            get_index_for_sharev2_search('bad-request')
```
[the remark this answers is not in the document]
aaxelb: the other two aren't errors anymore; they now return a default that depends on settings and feature flags

...should add tests for that, tho

trove/vocab/trove.py
```diff
@@ -235,7 +235,7 @@ def trove_browse_link(iri: str):
             TROVE.valueSearchFilter,
             TROVE.pageSize,
             TROVE.pageCursor,
-            TROVE.sort,
+            # TROVE.sort,
```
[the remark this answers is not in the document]
aaxelb: never did, tbh -- as implemented, has always ignored any `sort` param and sorted from most- to least-used, and changing that seemed like it'd raise more questions and be more difficult than correcting the docs

share/search/index_strategy/trovesearch_denorm.py
aaxelb: the `denorm` index has a document for the card itself and additional documents for each iri value in the card contents (with the `card` object duplicated across all of them, so a value-search can filter by anything on the card)
```
{card: ...}
{card: ..., iri_value: ...}
{card: ..., iri_value: ...}
{card: ..., iri_value: ...}
...
```
after indexed like that, if a subsequent update is missing any previous `iri_value` (say, a subject or contributor is removed), the corresponding value docs need to be deleted -- the easy/naive path was this `delete_by_query` to drop all value docs for the chunk of cards that are about to be indexed

(note it's not just for backfills -- the urgent queue is chunked as well, tho you'll more often get chunks of size one)

share/search/index_strategy/trovesearch_denorm.py
aaxelb: it is indeed same pagination as in `flats` -- "terribly inefficient" was a label i applied while coding (and copy-pasted here), but in practice it's not _too_ much slower to load N pages and discard N-1 of them (since N is small in practice)

maybe "terribly hacky" is a better descriptor

share/search/index_strategy/trovesearch_denorm.py
aaxelb: my intent was/is to delete `trove_indexcard_flats` as soon as `trovesearch_denorm` is in stable use, so tried to avoid putting much effort into updating it further -- but agree there are a few things (like cursors) that would make more sense reused than duplicated, yeah

mfraezz (approved):

PR #829, opened 2024-10-31
[ENG-6508] fix: two few adapters

aaxelb:
smooth out some hacks to work with slightly updated `django-allauth`

(in allauth, "oauth2 adapters" and "social account adapters" are different things with overlapping name that (by happenstance, previously) were non-overlapping enough they could be merged into a single subclass without much problem -- alas, no longer)


mfraezz (approved):
LGTM. Would be nice if there were tests, but since this is effectively just configuring a third-party library, makes sense that there wouldn't be.

Pass complete :octocat:

PR #832, opened 2024-11-20
[ENG-6576] drop trailing slashes for value-search

aaxelb:
- when indexing iri values for value-search aggregations, drop trailing slash and ensure a single iri (not including synonyms)
- update `trovesearch_denorm` mappings: single iri for each value, better synonym handling
- update tests with fixes while leaving old _flats strategy as-is (because it'll be removed before long...)


mfraezz (commented):
Generally looks good, however this change will require a reindex, which also means it'd be a good time to implement sharding. Could that be added to this PR, or should it be separate?

Pass complete :octocat:

mfraezz (approved):

PR #834, opened 2024-12-02
[ENG-6284] render tsv/csv

aaxelb:
allow rendering search responses as lines of tab-separated or comma-separated values

main point:
- add `simple_tsv` and `simple_csv` renderers in `trove.render`
  - can be seen with query param `acceptMediatype=text/tab-separated-values` or `acceptMediatype=text/csv`
  - get default columns from static `DEFAULT_TABULAR_SEARCH_COLUMN_PATHS` in `trove.vocab.osfmap`
- allow "download" responses -- add `withFileName=foo` query param to get a response with `Content-Disposition: attachment` and a filename based on "foo"
- allow absurd page sizes

changes made along the way:
  - introduce `ProtoRendering` as renderer output type, to better decouple rendering from view logic
    - include `StreamableRendering` for responses that might could be streamed, like csv/tsv (tho it's not currently handled any differently from `SimpleRendering`)
    - reshape `BaseRenderer` (and each existing renderer) to have a consistent call signature (and return `ProtoRendering`)
      - replace `trove.render.get_renderer` with `trove.render.get_renderer_type` -- instantiate the renderer with response data
  - add `trove.views._responder` with common logic for building a django `HttpResponse` for a `ProtoRendering`
    - consistently handles `withFileName`/`Content-Disposition`
  - move some osf-specific constants to `trove.vocab.osfmap` for easier reuse
  - pull out some abstractable logic:
    - from existing `trove.render.simple_json` into `trove.render._simple_trovesearch` (for renderers that include only the list of search results)
    - from existing `tests.trove.derive._base` into `tests.trove._input_output_tests` (for tests following the same simple input/output pattern as deriver and renderer tests)
  - add `tests.trove.render` to cover the new renderers `simple_tsv` and `simple_csv`, as well as the existing renderers `jsonapi`, `simple_json`, `jsonld`, and `turtle`
    - minimally update existing renderers to create consistent output


trove/render/simple_tsv.py
```diff
@@ -0,0 +1,143 @@
...
+Jsonpath = typing.Iterable[str]  # path of json keys
+
+_MULTIVALUE_DELIMITER = ' ; '  # possible improvement: smarter in-value delimiting?
+_VALUE_KEY_PREFERENCE = ('@value', '@id', 'name', 'prefLabel', 'label')
+
+
+class TrovesearchSimpleTsvRenderer(SimpleTrovesearchRenderer):
+    MEDIATYPE = mediatypes.TAB_SEPARATED_VALUES
+    INDEXCARD_DERIVER_IRI = TROVE['derive/osfmap_json']
+    CSV_DIALECT: type[csv.Dialect] = csv.excel_tab
```
mfraezz, line 23: Nit: Given this, I'd expect the inheritance between CSV/TSV Renderers to be swapped, but it's not important

trove/trovesearch/page_cursor.py
```diff
@@ -14,10 +14,13 @@
 MANY_MORE = -1
 MAX_OFFSET = 9997
 
+DEFAULT_PAGE_SIZE = 13
+MAX_PAGE_SIZE = 10000
```
mfraezz, line 18: Is this maximum reasonable? Looks like it was previously `101`

mfraezz (approved):
A couple nits or questions, but nothing blocking. Tests look sufficient, but behavior should still be confirmed manually on staging. 

Pass complete :octocat:

trove/render/html_browse.py
```diff
@@ -22,42 +24,58 @@
...
 
 STABLE_MEDIATYPES = (mediatypes.JSONAPI,)
-UNSTABLE_MEDIATYPES = (mediatypes.TURTLE, mediatypes.JSONLD, mediatypes.JSON,)
+UNSTABLE_MEDIATYPES = (
+    mediatypes.TURTLE,
+    mediatypes.JSONLD,
+    # TODO: below are only for search/index-card views
+    mediatypes.JSON,
+    mediatypes.TAB_SEPARATED_VALUES,
+    mediatypes.COMMA_SEPARATED_VALUES,
```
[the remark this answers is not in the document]
aaxelb: oh, because the mediatype for tsv is `text/tab-separated-values`, while the mediatype for comma-separated values is `text/csv`, and at different times i played with that in different ways -- probably should embrace "tsv"/"csv" for consistency, tho, true

trove/render/html_browse.py
```diff
@@ -22,42 +24,58 @@
...
+                'href': staticfiles_storage.url('css/browse.css'),
+            })
+        _body_attrs = {
+            'class': 'BrowseWrapper',
+            'style': f'--random-turn: {random.random()}turn;',
+        }
+        with self.__nest('body', attrs=_body_attrs):
+            self.__render_subj(self.focus_iri),
+            self.__render_mediatype_links()
+            # TODO: <details> with unvisited triples in self.data (unreachable from focus_iri)
```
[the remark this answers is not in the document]
aaxelb: hmm nope, was a possible improvement to a feature that no one asked for, to handle a case we're not using -- will change to a `# possible improvement: ...`

trove/render/simple_tsv.py
aaxelb: yeahh i elevated tsv on purpose to emphasize tabs as a preferred delimiter over commas, treating csv as back-compat, but then the standard python import for either is `csv` and naming got confused from there -- agree it would make more sense the other way

trove/trovesearch/page_cursor.py
aaxelb: update: now streams, loading only one page (~100 rows) at a time, but streaming more than ~4000 items total still times out -- can further optimize or we can talk about increasing those timeouts for responses that are actively sending data...

mfraezz (approved):
LGTM :octocat:

trove/trovesearch/page_cursor.py
aaxelb: am seeing timeouts at just after 30sec on production
Screenshot 2025-01-02 at 10 38 01

looking at the `.csv.part` files, each one managed to stream a few pages before it dies... unsure what's stopping it

PR #836, opened 2024-12-03
[ENG-6654] fixfix: correct misunderstanding, handle conflicts

aaxelb:
- partly revert #835 (remove `ensure_ack`, `ack_callback`) -- messages can _only_ be ack'd thru the channel they were received by, so trying to recover a channel to ack thru does nothing
- handle easily-detectable delete_by_query conflicts in a followup celery task (so the indexer doesn't fall over on it)
- local setup: allow the worker's `daemon` user to access elastic8
- update consumer prefetch limit to match the indexing chunk size (seems to avoid connection errors, [maybe this is why](https://stackoverflow.com/a/36114095))
- try avoiding connection errors when calling `message.ack()` -- do a bit more book-keeping to know when a bulk elastic action is the last for its index-card, so it can ack immediately then instead of waiting 'til the whole chunk has completed
- test updates to handle changes...

### DIFF
--- a/share/search/daemon.py
+++ b/share/search/daemon.py
@@ -8,7 +8,6 @@ import random
 import threading
 import time
 
-import amqp.exceptions
 from django.conf import settings
 import kombu
 from kombu.mixins import ConsumerMixin
@@ -61,8 +60,6 @@ class IndexerDaemonControl:
             index_strategy=index_strategy,
             message_callback=_daemon.on_message,
         )
-        # give the daemon a more robust callback for ack-ing
-        _daemon.ack_callback = _consumer.ensure_ack
         # spin up daemonthreads, ready for messages
         self._daemonthreads.extend(_daemon.start())
         # start a thread to consume messages from this strategy's queues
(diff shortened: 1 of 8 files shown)


share/search/daemon.py
```diff
@@ -82,7 +79,7 @@ def stop_daemonthreads(self, *, wait=False):
 
 
 class KombuMessageConsumer(ConsumerMixin):
-    PREFETCH_COUNT = 7500
+    PREFETCH_COUNT = settings.ELASTICSEARCH['CHUNK_SIZE']
```
mfraezz, line 82: It looks like this is currently set to 250 on prod. Should it be higher?

share/search/index_strategy/trovesearch_denorm.py
```diff
@@ -154,15 +155,14 @@ def _paths_and_values_mappings(self):
...
-                {'range': {'chunk_timestamp': {'lt': messages_chunk.timestamp}}},
-            ]}},
+        task__delete_iri_value_scraps.apply_async(
+            kwargs={
+                'index_strategy_name': self.name,
+                'indexnames': list(indexnames),
+                'card_pks': messages_chunk.target_ids_chunk,
+                'timestamp': messages_chunk.timestamp,
+            },
+            countdown=3,  # TODO: config?
```
mfraezz, line 165: Minor: Should probably be configurable, yes.
Something like 
`settings.ELASTICSEARCH['COUNTDOWN_DELAY'] = int(os.environ.get('ELASTICSEARCH_COUNTDOWN_DELAY', 3))`
would probably be reasonable

mfraezz (approved):
One Q, one minor thing to mitigate TODOs, but generally looks good.

Pass complete :octocat:

share/search/daemon.py
aaxelb: might be the default should be lower! maybe worth experimenting during a backfill

mfraezz: We can keep it at 250 and see how it behaves. If the issue was that rabbit was timing out prefetched items on partial-chunk acks, having them be equal sizes should mitigate for most sensible values.

share/search/index_strategy/trovesearch_denorm.py
aaxelb: went with `ELASTICSEARCH_POST_INDEX_DELAY`

PR #837, opened 2024-12-05
[ENG-5907] osfmap: suggest `affiliation` for preprints

aaxelb:
when feature flag PREPRINT_AFFILIATIONS is up, suggest `affiliation` instead of `creator.affiliation` for searching preprints


mfraezz (approved):
LGTM ⭐

PR #839, opened 2025-01-15
Fixed local setup

opaduchak:


aaxelb (approved):
oh good catch... i guess it makes sense that chown-ing the volume mount might cause problems...

PR #842, opened 2025-02-18
[ENG-6775] Added description to institution response

bodintsov:
added description to institution response


bodintsov:
Added new field `description` and `link`  to the property `propertyPath` to institutions' response.

PR #843, opened 2025-02-21
[ENG-6775] Refactored link and link text as separate variables

bodintsov:
Ticket https://openscience.atlassian.net/browse/ENG-6775


aaxelb (approved):
👍

PR #846, opened 2025-02-28
more index wrangling...

aaxelb:
- reduce indexing of unnecessary literal-text fields
- add confirmation checkbox to delete buttons at `/admin/search-indexes` (and allow deleting more things)


felliott (approved):
Looks good!

really really

PR #847, opened 2025-03-05
[gizmob] share clean(up)grade [ENG-7388][ENG-7389][ENG-7390]

aaxelb:
- delete a bunch of sharev2 code (and corresponding tests) that we don't need/use anymore
- update existing code/tests to be ok with those deletions


felliott (approved):
Yay! ✂️ ⚔️ 🔪  👍

PR #852, opened 2025-03-26
[ENG-7411] poetry and assorted cleanup

aaxelb:
- use `poetry` to install/manage dependencies
  - remove `requirements.txt`, `dev-requirements.txt`, `constraints.txt`
  - add dependencies to `pyproject.toml`
  - add `poetry.lock`
  - update `Dockerfile` and `docker-compose.yml` to use `poetry`
- assorted cleanup
  - update github action to use python 3.13
  - remove (now unused) elasticsearch5 code (including `sharev2_elastic5` index strategy)
  - remove (now unused) `trove_indexcard_flats` index strategy


Dockerfile
```diff
@@ -46,14 +60,14 @@ ARG GIT_COMMIT=
...
 ### Dist
 FROM app AS dist
 
+RUN $POETRY_HOME/bin/poetry install --compile --only deploy
+
 ### Dev
 FROM app AS dev
 
-RUN pip install --no-cache-dir -c /code/constraints.txt -r /code/dev-requirements.txt
+RUN $POETRY_HOME/bin/poetry install --compile --only dev
```
felliott, line 73: Looks good. Docker build had a few nitpicks, but these aren't dealbreakers:

```
 3 warnings found (use docker --debug to expand):                                                                                                                                                           
 - FromAsCasing: 'as' and 'FROM' keywords' casing do not match (line 1)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 60)
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 61)
```

.github/workflows/run_tests.yml
```diff
@@ -54,32 +48,28 @@ jobs:
       - name: install non-py dependencies
         run: sudo apt-get update && sudo apt-get install -y libxml2-dev libxslt1-dev libpq-dev git gcc
 
-      - name: set up python${{ matrix.python-version }}
+      - name: Install poetry
+        run: pipx install poetry
```
felliott, line 52: Should this be pinned to the same version as the Dockerfile (2.1.1)?  I don't have a strong opinion. It might be worth to leave unpinned as a canary.

felliott (approved):
Looks good to me!  I had a couple of small questions, but neither is a dealbreaker.

Cheers,
@felliott

PR #855, opened 2025-04-21
[ENG-7455] easy bulk-delete long-past harvest

aaxelb:
- add `delete_pretrove_data` management command that deletes every `RawDatum` associated with a "pre-trove" `SourceConfig` (anything not ingested via the rdf-based `/trove/` api)
- update `from_raw_datum` FK (on `IndexcardRdf`-based models) with `on_delete=DO_NOTHING` to avoid django's emulated cascade-delete -- may error if trying to delete data still referenced

[ENG-7455]

[ENG-7455]: https://openscience.atlassian.net/browse/ENG-7455?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


felliott (approved):
LGTM, but I don't have the deepest understanding of DjORM. Everything reads as sane, tho.

PR #868, opened 2025-05-09
[ENG-7445] Added IndexcardJsonDeriver

bodintsov:
Added IndexcardJsonDeriver

https://openscience.atlassian.net/browse/ENG-7445

### DIFF
--- a/tests/trove/derive/test_osfmap_json.py
+++ b/tests/trove/derive/test_osfmap_json.py
@@ -1,548 +1,557 @@
 import json
 from trove.derive.osfmap_json import OsfmapJsonDeriver
+from trove.derive.osfmap_json_mini import IndexcardJsonDeriver
 
 from ._base import BaseIndexcardDeriverTest
 
+expected_outputs = {
+    'blarg-item': {
+        "@id": "blarg:my_item",
+        "resourceType": [{"@id": "blarg:Item"}],
+        "title": [{
+            "@value": "title",
+            "@language": "en"
+        }],
+        "creator": [{
+            "@id": "blarg:me",
+            "resourceType": [{"@id": "Person"}],
+            "name": [{"@value": "me me"}]
+        }],
+        "dateCreated": [{"@value": "2024-02-14"}],
+    },
+    'blarg-project': {
+        "@id": "blarg:my_project",
+        "resourceType": [
+            {"@id": "Project"},
+            {"@id": "blarg:Item"},
+        ],
+        "title": [{
+            "@value": "title",
+            "@language": "en",
+        }],
+        "creator": [{
+            "@id": "blarg:me",
+            "resourceType": [{"@id": "Person"}],
+            "name": [{"@value": "me me"}]
+        }],
+        "dateCreated": [{"@value": "2024-02-14"}],
+    },
+    'sharev2-with-subjects': {
+        "@id": "http://osf.example/chair/",
+        "resourceType": [
+            {"@id": "sharev2:CreativeWork"},
+            {"@id": "sharev2:Publication"},
+            {"@id": "sharev2:Registration"},
+        ],
+        "conformsTo": [{"name": [{"@value": "Open-Ended Registration"}]}],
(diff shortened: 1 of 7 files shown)


trove/derive/__init__.py
```diff
@@ -1,19 +1,17 @@
...
-    osfmap_json,
+    osfmap_json_mini,
     oaidc_xml,
 )
 
-
 DERIVER_SET = (
     sharev2_elastic.ShareV2ElasticDeriver,
-    osfmap_json.OsfmapJsonDeriver,
+    osfmap_json_mini.IndexcardJsonDeriver,
```
aaxelb, line 9: hmm, want to reduce potential confusion with the "mini" fully replacing the original `osfmap_json`...

short of redesigning `trove.derive` to be less confusing, tho, should make sure the two have distinct `deriver_iri` -- could copy the [old deriver_iri](https://github.com/CenterForOpenScience/SHARE/blob/077710829626ea053af3b81d700b54f2f8e7fdcb/trove/derive/osfmap_json.py#L16-L19) to the new "mini" class, then update the old to return `TROVE['derive/osfmap_json_full']` instead (this way current api will start using the "mini", but the "full" has a distinct identifier in case it's ever used another way)

tests/trove/derive/test_osfmap_json.py
```diff
@@ -1,548 +1,559 @@
...
-                {"@value": "IA"},
-                {"@value": "IMLS"},
-                {"@value": "OSF"}
-            ]
-        },
-    }
+
+class TestIndexcardJsonDeriver(BaseIndexcardDeriverTest):
+    deriver_class = IndexcardJsonDeriver
+    expected_outputs = expected_outputs
```
aaxelb, line 556: alas, if these two derivers were supposed to have the same outputs we wouldn't need both of them -- add separate `expected_outputs` for the "mini" to check it is correctly pruned (which it's not, thanks to that `return True if ... else True`)

aaxelb (changes requested):

trove/derive/osfmap_json_mini.py
```diff
@@ -0,0 +1,61 @@
...
+    ns.OSFMAP.hostingInstitution,
+    ns.OSFMAP.keyword,
+    ns.OSFMAP.contains,
+    ns.OSFMAP.fileName,
+    ns.OSFMAP.filePath,
+    ns.OSFMAP.isContainedBy
+})
+
+
+class IndexcardJsonDeriver(OsfmapJsonFullDeriver):
```
aaxelb, line 39: (was going to ignore this but then i got confused often enough while reading it seems worthwhile to) rename `IndexcardJsonDeriver` to `OsfmapJsonMiniDeriver`, to be consistent with the module and parent class naming

tests/trove/derive/test_osfmap_json.py
```diff
@@ -544,5 +541,533 @@ def assert_outputs_equal(self, expected, actual):
...
             ]
-        },
+        }
     }
+
+    def assert_outputs_equal(self, expected, actual):
+        self.assertEqual(expected, json.loads(actual))
+
+
+class TestIndexcardJsonDeriver(BaseIndexcardDeriverTest):
```
aaxelb, line 551: belongs in its own file `test_osfmap_json_mini.py` (aligning with the other test files in this dir and mirroring code file layout)

aaxelb (changes requested):
small requests and a rename; seems near ready

tests/trove/derive/test_osfmap_json.py
bodintsov: maybe not a `test_osfmap_json_mini.py` but `test_indexcard_json_mini.py`?

aaxelb:
(oops meant to merge to develop -- have no done [by hand](https://github.com/CenterForOpenScience/SHARE/commit/4a73dfe723a34c51e16c0a85b7583393ca542490)

PR #869, opened 2025-05-13
[ENG-7446] Allow fields[*]

bodintsov:
Allow fields[*]

https://openscience.atlassian.net/browse/ENG-7446

### DIFF
--- a/trove/render/simple_csv.py
+++ b/trove/render/simple_csv.py
@@ -78,7 +78,7 @@ class TabularDoc:
         _pathlists: list[Iterable[Propertypath]] = []
         if self.trove_params is not None:  # hacks
             if GLOB_PATHSTEP in self.trove_params.attrpaths_by_type:
-                _pathlists.append(self.trove_params.attrpaths_by_type['*'])
+                _pathlists.append(self.trove_params.attrpaths_by_type[GLOB_PATHSTEP])
             if isinstance(self.trove_params, ValuesearchParams):
                 _expected_card_types = set(self.trove_params.valuesearch_type_iris())
             elif isinstance(self.trove_params, CardsearchParams):
--- a/trove/render/simple_csv.py
+++ b/trove/render/simple_csv.py
@@ -13,7 +13,7 @@ from trove.trovesearch.search_params import (
     CardsearchParams,
     ValuesearchParams,
 )
-from trove.util.propertypath import Propertypath
+from trove.util.propertypath import Propertypath, GLOB_PATHSTEP
 from trove.vocab import mediatypes
 from trove.vocab import osfmap
 from trove.vocab.namespaces import TROVE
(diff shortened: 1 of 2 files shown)


trove/util/trove_params.py
```diff
@@ -107,13 +108,20 @@ def _gather_attrpaths(cls, queryparams: _qp.QueryparamDict, shorthand: rdf.IriSh
...
+                    )
                 else:
                     for _type in _qp.split_queryparam_value(_typenames):
                         _type_iri = shorthand.expand_iri(_type)
                         _requested[_type_iri].extend(
                             parse_propertypath(_path_value, shorthand)
                             for _path_value in _qp.split_queryparam_value(_param_value)
                         )
+            if wildcard_paths:
+                _requested['*'].extend(wildcard_paths)
```
aaxelb, line 124: hmm, should be able to add this logic without duplicating the `parse_propertypath`/`split_queryparam_value` comprehension (or adding a separate `wildcard_paths` list)
```py
                    for _type in _qp.split_queryparam_value(_typenames):
                        _type_key = (
                            GLOB_PATHSTEP
                            if _type == GLOB_PATHSTEP
                            else shorthand.expand_iri(_type)
                        )
                        _requested[_type_key].extend(...
```
(this is a bit more lenient; allows something like `fields[*,Foo]=...`, but that's probably ok)

trove/render/simple_csv.py
```diff
@@ -83,6 +83,8 @@ def _column_paths(self) -> Iterator[Propertypath]:
                 _expected_card_types = set(self.trove_params.cardsearch_type_iris())
             else:
                 _expected_card_types = set()
+            if '*' in self.trove_params.attrpaths_by_type:
+                _pathlists.append(self.trove_params.attrpaths_by_type['*'])
```
aaxelb, line 87: prefer `trove.util.propertypath.GLOB_PATHSTEP` over literal `'*'`

trove/render/simple_csv.py
aaxelb, line 1: (can't comment on the line directly but) should also update [_column_paths return value](https://github.com/CenterForOpenScience/SHARE/blob/921cb4bd13e276576d424867d879bb6da00acd26/trove/render/simple_csv.py#L97) to deduplicate columns (while preserving requested order), in case `fields[*]=...` is ever given alongside type-specific `fields[Foo]=...` -- maybe a quick util:
```py
def iter_unique(iterable):
    _seen = set()
    for _item in iterable:
        if _item not in _seen:
            _seen.add(_item)
            yield _item
```
and
```py
return iter_unique(itertools.chain.from_iterable(_pathlists))
```

trove/render/simple_csv.py
aaxelb, line 87: readability nit: this is inserted between where `_expected_card_types` is set and used, despite having nothing to do with that -- would flow better inserted at the start of the `if self.trove_params is not None:` block, leaving the more complicated type-specific logic uninterrupted

aaxelb (changes requested):

PR #881, opened 2025-07-28
[ENG-8409] fix: better handle errors in celery result backend

aaxelb:
- use celery's own "retry on result-backend error" logic
    - new env var: `CELERY_RESULT_BACKEND_MAX_RETRIES` (default 17)
    - retry on OperationalError (for connection problems and such)
      and IntegrityError (for conflicting `get_or_create` in overlapping
      transactions, for example)
    - move `die_on_unhandled` decorator to public methods, so it's only
      used after max retries
- port fixes from celery's `BaseKeyValueStoreBackend`:
    - avoid clobbering successes with "worker lost" or other errors
    - avoid error trying to get non-existent task results
- use celery's `BaseBackend` instead of `BaseDictBackend`
  (equivalent for back-compat; let's use the better name)


felliott (approved):
LGTM

